PR #108, opened 2022-06-23
Several Ethereum related fixes

### LINKED ISSUE #93: EVM Address of accounts with alias of ECDSA(secp256k1) public key incorrect
### Description

When an account has an alias that is an ECDSA(secp256k1) address the EVM address is reported incorrectly.  It reports the alias when it should be reporting a value that is a function of the alias.

The `evm_address` of the mirror node query reports this value, for example https://testnet.mirrornode.hedera.com/api/v1/accounts/0.0.45911647

If it needs to be calculated it's quite baroque.  Get the _uncompressed_ public key, drop the recovery id, then sha256k1 the x and y parameter, then take the 20 trailing bytes.

### Steps to reproduce

https://hashscan.io/#/testnet/account/0.0.45911647 - ERC-20 address is wrong, should be `a94f 5374 fce5 edbc 8e2a 8697 c153 3167 7e6e bf0b` but is `3a21 033a 5141 7646 6fa8 15ed 481f fad0 9110 a2d3 44f6 c9b7 8c1d 14af c351 c3a5 1be3 3d`, which is the account alias.

### Additional context

_No response_

### Hedera network

testnet

### Version

Build date: Thu, 09 Jun 2022 21:51:05 GMT

### Operating system

_No response_

Neurone:
**Description**:
This PR modifies mainly Account and Contract pages to apply several fixes:
* Removed references to `solidiy_address`. `solidity_address` does not exist anymore, it became `evm_address` (see hashgraph/hedera-mirror-node#3217)
* Fixed references to `evm_address` (#93). If `evm_address` is present, that is used
* Fixed Ethereum address computation. If `evm_address` is not present and the key is ECDSA, the address is computed with ethers.js. The old method was wrong because it was using the compressed public key to create the address
* Fixed TypeScript linter's warnings and errors
* Fixed README.md to follow Markdown linter's suggestions

This PR also introduce few improvements:
* Alias is now human-readable, as intended in [HIP-32](https://hips.hedera.com/hip/hip-32)
* Improved some wording (i.e. _ERC20 Address_ becomes _Ethereum Address_, _Key_ becomes _Admin Key_)

**Related issue(s)**:

Fixes #93 

**Checklist**


ericleponner:
In EntityCache class, I see that you want to change:
    let timeoutID = -1
as 
   let timeoutID: null | ReturnType<typeof setTimeout> = null;

Why ? 
According MDN, setTimeout() returns a number.
When this number is positive, it’s a valid timer.
So any negative number can be used for no value.

Neurone:
> In EntityCache class, I see that you want to change: let timeoutID = -1 as let timeoutID: null | ReturnType = null;
> 
> Why ? According MDN, setTimeout() returns a number. When this number is positive, it’s a valid timer. So any negative number can be used for no value.

With the introduction of `ethers.js`, the compiler started complaining about `setTimeout` function:

```shell
TS2322: Type 'Timeout' is not assignable to type 'number'.
```

This is because it tries to use the NodeJS's setTimeout function:

```TypeScript
setTimeout(callback: (...args: any[]) => void, ms: number, ...args: any[]): NodeJS.Timer;
```

and not the DOM's one:

```TypeScript
setTimeout(handler: (...args: any[]) => void, timeout: number): number;
```

I updated the code that was using `setTimeout` to reflect that change but, if you prefer, I can try to find a way to force using the DOM version.

ericleponner:
Ok, I see.
So may be you could simply use window.setTimeout() and window.clearTimeout() for now.
Explorer code is not designed to be run outside of the navigator.

ericleponner:
Concerning account alias, I see that you want to:
1) remove base32 decoding and display the raw Base32 encoding
2) move alias before key field and thus give more visibility to it
Any arguments for this ?

Neurone:
> Concerning account alias, I see that you want to:
> 
> 1. remove base32 decoding and display the raw Base32 encoding
> 2. move alias before key field and thus give more visibility to it
> 
> Any arguments for this ?

Hi Eric, sure!

1) Reading [HIP-32](https://hips.hedera.com/hip/hip-32) I understand alias are meant to be used by humans - like account ID - and they should be presented and used with base32 encoding. HIP-32 shows `0.0.CIQNOWUYAGBLCCVX2VF75U6JMQDTUDXBOLZ5VJRDEWXQEGTI64DVCGQ` as a valid alias.

Furthermore, that value is the same returned by mirror node API, and I think it's useful when an explorer offers at least the same information it gets from the API, plus maybe other info, but not less.

About aliases in general, I also plan to send some PRs to the JavaScript SDK for the same reason: it currently supports sending txs to `0.0.302d300706052b8104000a03220003c7f5360add55a5237b530ed4720a99ad9e52247e8a8e5181f19dfebe26e97608` but not `0.0.CIQNOWUYAGBLCCVX2VF75U6JMQDTUDXBOLZ5VJRDEWXQEGTI64DVCGQ`. I think this approach is also taken by the other SDKs, and I think it should be changed to follow HIP-32 correctly.

2) About moving alias up in the page, I think that alias is something you need to be aware of because it's very similar to a nickname for your account ID. And also it's a long value, and it does not fit in the table below ^^

ericleponner:
In AccountDetails.vue:
`<div class="h-is-text-size-3"><StringValue :string-value="account?.alias"/></div>`
should be
`<StringValue :string-value="account?.alias »/>`
to respect detail page styling.

I guess base32 value will not look very nice in that case...
An alternative layout should probably necessary then.

ericleponner:
> Hi Eric, sure!
> 
> 1. Reading [HIP-32](https://hips.hedera.com/hip/hip-32) I understand alias are meant to be used by humans - like account ID - and they should be presented and used with base32 encoding. HIP-32 shows `0.0.CIQNOWUYAGBLCCVX2VF75U6JMQDTUDXBOLZ5VJRDEWXQEGTI64DVCGQ` as a valid alias.
> ...
> 2. About moving alias up in the page, I think that alias is something you need to be aware of because it's very similar to a nickname for your account ID. And also it's a long value, and it does not fit in the table below ^^

Ok, I got your points, Guiseppe.
I’ll check with some UI folks and come back on this later.
Thanks

Neurone:
> I guess base32 value will not look very nice in that case... An alternative layout should probably necessary then.

You are right, I used `h-is-text-size-3` because of that ^^ Unfortunately, I'm not a good UI designer, so if you prefer I can use the styling you suggested until we find a prettier solution.

Neurone:
@ericleponner The PR is now fixed with your suggestions, resolved merge conflicts and a couple of new fixes (the linter)

ericleponner (dismissed):
@Neurone Thanks for your contribution.
Regarding alias field, the current UI is temporary.
We’ll revisit it soon.

Neurone:
> @Neurone : merge is blocked because your commits are not signed ; signing is mandatory. You need to signed all your commits and resubmit. There are some instructions here: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

I'm working on it :)

ericleponner:
@Neurone : sorry but I think signing is still not ok; GitHub should display a <verified> tag next to your commit. Yes ... it's really a pain ... I know...

See:
https://docs.github.com/en/authentication/troubleshooting-commit-signature-verification/checking-your-commit-and-tag-signature-verification-status

Neurone:
> @Neurone : sorry but I think signing is still not ok; GitHub should display a tag next to your commit. Yes ... it's really a pain ... I know...
> 
> See: https://docs.github.com/en/authentication/troubleshooting-commit-signature-verification/checking-your-commit-and-tag-signature-verification-status

Oh, now I got it ^^ you are talking about PGP signatures, I thought you were referring to the `signed-off-by` tag at the end of each commit, sorry. I'll fix that.

Neurone:
@ericleponner It should be ok now.

ericleponner (approved):

ericleponner:
@Neurone : thanks Giuseppe.

PR #171, opened 2022-08-05
Show build version (fixes #137)

### LINKED ISSUE #137: Display the build version, not just the build date/time
### Problem

When submitting tickets, it requests that I have the build version number but https://hashscan.io/ only displays the build date/time.

### Solution

Can we display the build version, next to the build date/time?

### Alternatives

None

Neurone:
**Description**:

This PR show the build number using the package version present in the `package.json` at the moment of the build.

**Related issue(s)**:

Fixes #137 

**Notes for reviewer**:

N/A

**Checklist**


Neurone:
Hi @svienot, I see your point, and as you said, there's already a SI in place that works as expected. I would add nonetheless that the purpose of the build version is also into checking at a glance if two deployments are the same, regardless of the build time. I can use it to compare if my local build is the same of hashscan.io or if staging, testing and production env are the same, etc.

To remove the burden to manage a release version, and a corresponding release/feature detail page, we can go with the git revision of the commit that generated the build. Generation can be easily computed and automated, and it doesn't add management tasks. What do you think?

Neurone:
I try again, this time using git commit as build release ;) What do you think?

Neurone:
Replaced by #604

PR #172, opened 2022-08-06
Updated Code label

Neurone:
**Description**:

This PR modifies the _Code_ label to _Initcode_ (thanks @steven-sheehy).

Advantages:

* _Code_ can be mistaken by the actual deployed bytecode. Currently, you cannot recover the deployed bytecode via REST API. That bytecode is the actual [`initcode`](https://github.com/hashgraph/hedera-protobufs/blob/71e0b20b25aee4369c19c66eacdf5a97e5ca61bd/streams/contract_bytecode.proto#L39)
* We use consistent terminology across Hedera's applications

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/src/pages/ContractDetails.vue
+++ b/src/pages/ContractDetails.vue
@@ -92,7 +92,7 @@
               </template>
             </Property>
             <Property id="code">
-              <template v-slot:name>Code</template>
+              <template v-slot:name>Initcode</template>
               <template v-slot:value>
                 <ByteCodeValue :byte-code="contract?.bytecode"/>
               </template>


src/pages/ContractDetails.vue
```diff
@@ -92,7 +92,7 @@
               </template>
             </Property>
             <Property id="code">
-              <template v-slot:name>Code</template>
+              <template v-slot:name>Contract Creation Code</template>
```
steven-sheehy, line 95: Would prefer it called `Initcode` to align with both [HAPI](https://github.com/hashgraph/hedera-protobufs/search?q=initcode) and Ethereum standards as called out by Danno in a [HIP discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/513#discussion_r899646293). Hedera should have consistent terminology across its applications. We would rename the REST API field if we could.

steven-sheehy (changes requested):

src/pages/ContractDetails.vue
Neurone: I agree, just maybe `Initcode` is a bit too much _developer-oriented_ for an explorer web interface. What do you think about `Initialization Code`?

steven-sheehy: I think any one looking at hex data on a UI is probably a developer ;). Still prefer `Initcode` so everyone is in alignment. Perhaps a tooltip can clarify further.

svienot (approved):

steven-sheehy (approved):

PR #244, opened 2022-10-02
00135 show create transaction

### LINKED ISSUE #135: View account which created the account I'm viewing
### Problem

Looking at an account, I cannot presently find the transaction that created it.

### Solution

1. When viewing an account, add a field that indicates which account created the account I'm viewing.

### Alternatives

None

svienot:
**Description**:

With these changes, the new `created-timestamp` property of `AccountInfo` is used to query the corresponding `ACCOUNT CREATE` or `CONTRACT CREATE` transaction. To that end, a new `TransactionByTimestampLoader` is introduced.

**Related issue(s)**:

Fixes #135 

**Notes for reviewer**:

This only works for now on a local mirror node. Otherwise (in the absence of `created-timestamp` property) we display `None`. 

**Checklist**

- [X] Tested (unit, integration, etc.)

### DIFF
--- a/src/components/account/AccountLoader.ts
+++ b/src/components/account/AccountLoader.ts
@@ -44,6 +44,8 @@ export class AccountLoader extends EntityLoader<AccountBalanceTransactions> {
 
     public readonly balance: Ref<number|null> = computed(() => this.entity.value?.balance?.balance ?? null)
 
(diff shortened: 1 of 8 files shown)


svienot:
The recently added test inside TransactionNavigation.spec passed locally but failed on the server. So I skip it for now until further investigation.

ericleponner (approved):

Issue #304, opened 2022-11-06
Publish docker images


steven-sheehy:
### Problem

As a developer, I'd like to be able to run the explorer locally in concert with local consensus and mirror nodes. A prerequisite for working on #282.

### Solution

* Add a GitHub workflow that publishes a multi-architecture (arm64/amd64) Docker image to `gcr.io/hedera-registry/hedera-explorer` (See mirror node [example](https://github.com/hashgraph/hedera-mirror-node/blob/01bcf5817d91922035d166761ca6e3d874bb267f/.github/workflows/release-production.yml#L56))
* Tag version and a static `latest` tag
  * Since there are no git tags or releases, the version tag could use [calver](https://calver.org/) or some other incrementing date/time based versioning scheme

svienot:
As requested by Nathan, assigning the issue to him.

PR #306, opened 2022-11-07
Add a helm chart

### LINKED ISSUE #282: Write a Helm chart for hedera-mirror-node-explorer
### Problem

Coming from an initiative across hashgraph products(repos) we want to run on a cloud managed Kubernetes (GKE, EKE, AKS) that ops creates and manages.
Adding helm chart support has been logged in: https://github.com/hashgraph/hedera-services/issues/4068 
as well as in https://github.com/hashgraph/hedera-local-node/issues/189 
The mirror-node and relay already have this support. They can be referenced as examples.

### Solution

Write a [Helm](https://helm.sh/) chart for local node and manage the deployments.

### Alternatives

_No response_

steven-sheehy:
**Description**:

* Add a Helm chart for Kubernetes and future use in local node
* Change `Dockerfile` to be more appropriate for containers:
  * Change logs to write to stdout and stderr instead of files
  * Change nginx to run as non-root user
  * Change pid path to consolidated cache folder for easier volume mounting
  * Change port from 80 to 9090 so it can run as non-root

**Related issue(s)**:

Fixes #282

**Notes for reviewer**:

**Checklist**


chart/values.yaml
```diff
@@ -0,0 +1,148 @@
...
+
+global:
+  namespaceOverride: ""
+
+image:
+  pullPolicy: IfNotPresent
+  pullSecrets: []
+  registry: docker.io
+  repository: ssheehy/hedera-explorer
+  tag: latest  # Defaults to the chart's app version
```
steven-sheehy, line 50: This is using a static image from my personal Docker hub for now until #304 is implemented.

chart/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+appVersion: 0.1.0-SNAPSHOT
```
steven-sheehy, line 2: There's not a good appVersion to put here. This version will need to be updated on every release. Also, this chart will need to be published somewhere for consumption like GitHub pages or an artifact registry. These can be followups.

Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY ./ .
 RUN npm run build
 
 FROM nginx as production-stage
-RUN mkdir /app
+EXPOSE 9090
```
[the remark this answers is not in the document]
nathanklick: In fact IANA now recognizes 8080 as an alternate HTTP port: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=8080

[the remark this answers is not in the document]
steven-sheehy: I agree, but was following the standard port used throughout the code and didn't want it to be inconsistent. Also not sure if it would break things. If Simon and Eric are okay with changing to 8080 I can make that change.

chart/values.yaml
nathanklick: There will be after PR #312 ---> `gcr.io/hedera-registry/hedera-mirror-node-explorer:latest`

Dockerfile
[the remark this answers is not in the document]
nathanklick: I used 8080 in the docker image publish PR:

- #312

nginx-docker.conf
```diff
@@ -0,0 +1,28 @@
...
+                    '"$http_user_agent" "$http_x_forwarded_for"';
+  access_log /dev/stdout main;
+  sendfile        on;
+  keepalive_timeout  65;
+  server {
+    listen       9090;
+    server_name  localhost;
+    location / {
+      root   /app;
+      index  index.html;
```
svienot, line 21: index  index.html;
       try_files $uri $uri/ /index.html;

svienot, line 21:
```suggestion
      index  index.html;
      try_files $uri $uri/ /index.html;
```

Dockerfile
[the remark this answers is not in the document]
svienot: All good with 8080!

nginx-docker.conf
svienot: The catch-all redirect is needed due to Hashscan's single page app URL structure.

chart/values.yaml
nathanklick: Should be able to go ahead and update as above.

svienot:
@steven-sheehy I have merged this with main and adapted to use the docker image released on gcr.io
I have tested this locally with minikube. It works but when I execute helm upgrade --install hedera-explorer chart/
 helm does not take into account the newest explorer docker image available (like if caching the initial image found).
 Any clue ?

steven-sheehy:
Have to set image pull policy to Always. It's not the default because in production it's bad practice to overwrite existing images.

nginx.conf
steven-sheehy, line 1: Is this file needed by non-helm deploys?

chart/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+appVersion: latest
```
steven-sheehy, line 2: This is the latest image in GCR.

```suggestion
appVersion: 0.2.0
```

chart/values.yaml
```diff
@@ -0,0 +1,148 @@
...
+
+global:
+  namespaceOverride: ""
+
+image:
+  pullPolicy: IfNotPresent
+  pullSecrets: []
+  registry: gcr.io
+  repository: hedera-registry/hedera-mirror-node-explorer
+  tag: latest  # Defaults to the chart's app version
```
steven-sheehy, line 50: It's bad practice to use a static tag. Should leave blank so it uses the chart appVersion, which itself should be kept up to date.

```suggestion
  tag: ""  # Defaults to the chart's app version
```

README.md
```diff
@@ -119,6 +129,8 @@ To enable the Staking page and menu item, set the following variable to *true* i
 VUE_APP_ENABLE_STAKING=true
 ```
 
+### Customize configuration
+
```
steven-sheehy, line 133: Missing content?

Dockerfile
steven-sheehy, line 1: Why do we have an empty file? And why are we using a non-standard `CI.Dockerfile` on main? IMO, we should delete this file and rename the one on main to `Dockerfile`.

steven-sheehy (commented):
Thanks for picking this up. Had some questions/suggestions.

chart/values.yaml
```diff
@@ -0,0 +1,148 @@
...
+    path: /
+    port: http
+  initialDelaySeconds: 30
+  timeoutSeconds: 2
+
+replicas: 1
+
+resources:
+  limits:
+    cpu: 500m
```
steven-sheehy, line 105: Since Node.js is a (mostly) single-threaded process, we should bump this up to one CPU. Anything less than one CPU is inefficient with the way Node.js works. There are some other minor threads in Node.js so I would say put limit as:
```suggestion
    cpu: 1200m
```

chart/values.yaml
```diff
@@ -0,0 +1,148 @@
...
+  seccompProfile:
+    type: RuntimeDefault
+
+priorityClassName: ""
+
+readinessProbe:
+  httpGet:
+    path: /
+    port: http
+  initialDelaySeconds: 30
```
steven-sheehy, line 98: Might consider adjusting the initial delay for both probes lower since I think your app starts up really quick as it has no startup dependencies like a db.

nginx.conf
svienot: nginx-docker.conf is used only for Docker/Helm deploys.
it replaces nginx.conf, now deleted.

Dockerfile
svienot: Dockerfile was meant to be deleted. 
Now following your suggestion to rename CI.Dockerfile into Dockerfile.

nginx-docker.conf
```diff
@@ -0,0 +1,28 @@
...
+  include       /etc/nginx/mime.types;
+  default_type  application/octet-stream;
+  log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                    '$status $body_bytes_sent "$http_referer" '
+                    '"$http_user_agent" "$http_x_forwarded_for"';
+  access_log /dev/stdout main;
+  sendfile        on;
+  keepalive_timeout  65;
+  server {
+    listen       9090;
```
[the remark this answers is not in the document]
svienot: Now using port 8080.

README.md
svienot: No, the diff is a bit misleading because we have adjusted the structure of the README between the creation of this PR and now.

chart/values.yaml
svienot: Adjusting to 3 sec instead of 30 per your suggestion.

chart/Chart.yaml
svienot: appVersion has been updated.

svienot:
Thanks for your review @steven-sheehy. Hopefully, I have addressed all points.

steven-sheehy (commented):
LGTM

Issue #420, opened 2023-01-23
explorer shows contract has expired and entered grace period


mshakeg:
### Description

The explorer shows the warning **"Contract has expired and is in grace period"**  for number of contracts that have more than enough HBAR to pay their contract rent such as the following:
* https://hashscan.io/mainnet/contract/0.0.1060875
* https://hashscan.io/mainnet/contract/0.0.1060864

AFAIK contract rent is scheduled for sometime in February, so I don't think the above warning makes sense to show at this point in time.

Also, I think it would be helpful for the contract rent cost to be displayed on the explorer's contract page.
```
Contract Rent = Auto-renewal Cost + Storage(Key-Value) Costs
```

### Steps to reproduce

The above example contracts are sufficient evidence of the issue.

### Hedera network

mainnet

### Version

mirror-node@0.71.0

### Operating system

None

svienot:
This banner should not be displayed indeed on mainnet for the time being.
 As far as showing the contract rent cost, not sure yet which information will be available to HashScan.

svienot:
We will enable this capability soon. In the interim we can educate the users, if they so inquire. For now, we do nothing.

mshakeg:
@svienot I couldn't find an issue on the mirror node pertaining to contract storage so I created [hedera-mirror-node#5761](https://github.com/hashgraph/hedera-mirror-node/issues/5761)

PR #431, opened 2023-01-27
Hide 'Expires at' property on mainnet.

svienot:
**Description**:

This temporarily hides _Expires at_ property on AccountDetails and ContractDetails views, when running on mainnet.

**Related issue(s)**:

#420

**Checklist**


svienot:
From Mohsin Qamar:
  We will enable this capability soon. In the interim we can educate the users, if they so inquire. For now, we do nothing.

PR #568, opened 2023-04-26
Increased memory allocated to nodejs when running unit test

ericleponner:
**Description**:

This is an attempt to fix unit test failures observed on GitHub pipeline (npm error 134).


ericleponner:
Dead end : increasing heap size does not solve the issue.
Problem is solved by #573.
Closing

PR #611, opened 2023-05-25
Cleaning : reformat all *.vue files

ericleponner:
**Description**:

Changes below reformat all `*.vue` files.
This is the result  of
1)  IntelliJ `Reformat Code` command
2) a few adjustments (some `span` clauses were missing and causing test break)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -20,13 +20,13 @@
 
 <template>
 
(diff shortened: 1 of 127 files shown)


ericleponner:
No sure this reformatting changes are required. Might be a bug in IntelliJ 2023.1.x.
Closing…

PR #684, opened 2023-08-25
feat: Allow custom network config in helm chart

### LINKED ISSUE #683: The helm chart does not have a way to specify custom network config
### Problem

If I want to use localnet, I cannot specify a custom network-config.json

### Solution

Add a optional element 'customNetworkConfig' in values.yaml,
This would be used to create a configmap,
This configmap can be mounted as a file at /app/network-config.json

### Alternatives

_No response_

deepak-swirlds:
**Description**:

Allows us to specify customNetworkConfig in values.yaml
This allows adding any number custom network (like localnet)

example of values.yaml
```
config: |
  [ 
       ...
   {
        "name": "local",
        "displayName": "LOCALNET",
        "url": "/",
        "ledgerID": "03"
   }
 ]
```

**Related issue(s)**:

Closes #683 

**Notes for reviewer**:

**Checklist**


chart/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.customNetworkConfig }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: custom-networks-config
```
steven-sheehy, line 5:
```suggestion
  annotations: {{ toYaml .Values.annotations | nindent 4 }}
  labels: {{ include "hedera-explorer.labels" . | nindent 4 }}
  name: {{ include "hedera-explorer.fullname" . }}-config
  namespace: {{ include "hedera-explorer.namespace" . }}
```

chart/templates/deployment.yaml
```diff
@@ -61,4 +66,12 @@ spec:
         - name: {{ $name }}
         {{- tpl (toYaml $config) $ | nindent 10 }}
         {{- end }}
+        {{- if $.Values.customNetworkConfig }}
+        - name: custom-network-config-volume
+          configMap:
+            name: custom-networks-config
+            items:
+              - key: networks-config.json
+                path: networks-config.json
```
steven-sheehy, line 75: Shouldn't need to explicitly list all files. Also would be good to set more secure permissions.
```suggestion
            defaultMode: 420
```

chart/values.yaml
```diff
@@ -146,3 +146,32 @@ volumes:
   cache:
     emptyDir: {}
 
+# Enable this to use a custom network-config.json
+#customNetworkConfig: |
```
steven-sheehy, line 150: I think this should be just `config` and enabled by default. Also sort it within the file.
```suggestion
config: |
```

chart/templates/deployment.yaml
```diff
@@ -47,6 +47,11 @@ spec:
             - name: {{ $name }}
             {{- toYaml $config | nindent 14 }}
             {{- end }}
+            {{- if $.Values.customNetworkConfig }}
+            - name: custom-network-config-volume
+              mountPath: /app/networks-config.json
+              subPath: networks-config.json
+            {{- end }}
```
steven-sheehy, line 54: We already have custom volumes and volume mounts available from the values.yml that is templatized. I think we should just move this there so it's always present and has a default config with all public networks.

chart/templates/deployment.yaml
```diff
@@ -47,6 +47,11 @@ spec:
             - name: {{ $name }}
             {{- toYaml $config | nindent 14 }}
             {{- end }}
+            {{- if $.Values.customNetworkConfig }}
+            - name: custom-network-config-volume
+              mountPath: /app/networks-config.json
+              subPath: networks-config.json
```
steven-sheehy, line 53: Just mount all files in volume.
```suggestion```

File: chart/templates/deployment.yaml [hunk 6]
```diff
@@ -61,4 +66,12 @@ spec:
         - name: {{ $name }}
         {{- tpl (toYaml $config) $ | nindent 10 }}
         {{- end }}
+        {{- if $.Values.customNetworkConfig }}
+        - name: custom-network-config-volume
+          configMap:
+            name: custom-networks-config
```
steven-sheehy, line 72: Prefer a simpler name and need to templatize configMap name with previous recommendation.
```suggestion
        - name: config
          configMap:
            name: {{ include "hedera-explorer.fullname" . }}-config
```

steven-sheehy (changes requested):

chart/templates/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hedera-explorer-config
```
deepak-swirlds, line 4: since this name is used in values.yaml (where variables are not allowed)
I am not able to find a way to templatize this name

chart/templates/deployment.yaml
deepak-swirlds: configmap volume mounting is now moved to values.yaml

chart/values.yaml
```diff
@@ -140,9 +140,38 @@ updateStrategy:
...
   cache:
     mountPath: /var/cache/nginx
+  custom-networks-config:
+    mountPath: /app/networks-config.json
+    subPath: networks-config.json
 
 # Volume mounts to add to the container. The key is the volume name and the value is the volume definition. Evaluated as a template.
 volumes:
   cache:
     emptyDir: {}
```
steven-sheehy, line 151: Extra new line here between other volumes. Not sure if it causes problems in templating.

chart/values.yaml
```diff
@@ -24,8 +24,8 @@ autoscaling:
   minReplicas: 1
 
 # Environment variables to add to the container. The value can either be a string or an EnvVarSource. Evaluated as a template.
-env:
-  DOCKER_LOCAL_MIRROR_NODE_URL: "http://localhost:5551"
+#env:
```
steven-sheehy, line 27: Keep key so you don't get templating warnings

```suggestion
env: {}
```

chart/values.yaml
```diff
@@ -140,9 +140,38 @@ updateStrategy:
...
   cache:
     emptyDir: {}
 
+  custom-networks-config:
+    configMap:
+      name: hedera-explorer-config
+      
+
+config: |
+     [
```
steven-sheehy, line 158: I would've expected two spaces for indentation for the config but count 5. Not sure if it causes templating problems though.

chart/templates/configmap.yaml
steven-sheehy: You can templatize data in values.yml. See the mirror node chart for an [example](https://github.com/hashgraph/hedera-mirror-node/blob/bc14438df9dc329408a1a1a305a408d88b22cf19/charts/hedera-mirror-grpc/templates/deployment.yaml#L67).

steven-sheehy (changes requested):

steven-sheehy (dismissed):
LGTM

svienot:
The `appVersion` gets an old version of the Docker image (back in March).
Please update to the latest image version available: `23.7.0` 
Other than that, LG. I tested this locally on minikube.

PR #692, opened 2023-09-12
feat: add page for individual nft serials

Sheng-Long:
**Description**:
Add page for individual nft serial details (including recent transactions)

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Not sure how much token information should be retained/added for this

**Checklist**


src/components/token/NftHolderTable.vue
```diff
@@ -35,11 +35,13 @@
...
       @page-change="onPageChange"
 
-      :hoverable="false"
+      @cell-click="handleClick"
+      :hoverable="true"
       :narrowed="true"
       :striped="true"
       :mobile-breakpoint="ORUGA_MOBILE_BREAKPOINT"
 
+      style="cursor: pointer"
```
ericleponner, line 44: Is this line stil useful ?

src/components/token/NftHolderTable.vue
```diff
@@ -116,19 +135,10 @@ export default defineComponent({
...
-      ORUGA_MOBILE_BREAKPOINT
+      ORUGA_MOBILE_BREAKPOINT,
+      handleClick,
     }
   }
 });
 
 </script>
-
-<!-- --------------------------------------------------------------------------------------------------------------- -->
```
ericleponner, line 126: Right now, our convention is to keep comment banner with just <style/> tag.

svienot:
I played with the current stage of the implementation. Nice work.
Could you make the following adjustments:
- Recent Transactions Filter: the available types should be just those 7:
   {`TOKENMINT, CRYPTOAPPROVEALLOWANCE, CRYPTODELETEALLOWANCE, CRYPTOTRANSFER, TOKENWIPE, TOKENBURN, TOKENDELETION`}
   (we have checked this explicitly with Steven Sheehy, lead of mirror-node)
- Do not display properties when they are `null` (such as `Account ID`, `Delegating Spender`, `Spender ID`)
   We used to display a lot of "None" values, but we now try to move away from that and densify the contents as it gets more substantial
- _Nice to have_: when the metadata property is an IPFS hash, we could display a link (to the public IPFS gateway):
     e.g. if the metadata is: `ipfs://bafkreigds272du3nvdaywx4tgebkcw4uegxizku7cyhk3z6bhhwbnxsmwi`
     display a link to: [https://ipfs.io/ipfs/bafkreigds272du3nvdaywx4tgebkcw4uegxizku7cyhk3z6bhhwbnxsmwi](https://ipfs.io/ipfs/bafkreigds272du3nvdaywx4tgebkcw4uegxizku7cyhk3z6bhhwbnxsmwi)

svienot:
I suggest you adjust the failing TokenNavigation e2e test by replacing the sequence which looks for the now disappeared anchor inside the NFT Holder Table by something like this:
```
        cy.get('#nft-holder-table')
            .find('tbody tr')
            .should('be.visible')
            .should('have.length.at.least', 2)
            .eq(0)
            .find('td')
            .eq(0)
            .click()
            .then(($id) => {
                // cy.log('Selected account Id: ' + $id.text())
                cy.url().should('include', '/mainnet/token/' + nftId + '/' + $id.text())
                cy.contains('Serial Number ' + $id.text())
            })
```
and perhaps add a few assertions on the content of the page, such that we have at least a bit of e2e coverage for this brand new Serial Details page.

svienot (approved):

PR #754, opened 2023-11-13
fix: disable refreshes when navigating away from account details

### LINKED ISSUE #730: Update recent transactions in account flow
### Problem

I need to view the transactions related to the users Account. When I click into transactions, after I go to the users account,  and click the back button I get sent to the first page. I want to go back to the page that I was just recently on. This works as expected when navigating transactions from the Transactions tab, but does not work the same when navigating transactions from the Accounts tab.

### Solution

Fix the account level transaction navigation to match that of the Transactions tab. 

### Alternatives

_No response_

Sheng-Long:
**Description**:
Correctly unmount table controller and disable refreshes when it is unmounted

**Related issue(s)**:

Fixes #730 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/pages/AccountDetails.vue
+++ b/src/pages/AccountDetails.vue
@@ -387,7 +387,9 @@ export default defineComponent({
     })
     onBeforeUnmount(() => {
       mounted = false
-      transactionTableController.unmount()
+      if (transactionTableController.mounted.value) {
+        transactionTableController.unmount()
+      }
     })
     watch(accountId, () => {
       if (mounted) {
--- a/src/pages/AccountDetails.vue
+++ b/src/pages/AccountDetails.vue
@@ -387,9 +387,7 @@ export default defineComponent({
     })
     onBeforeUnmount(() => {
       mounted = false
-      if (accountId.value !== null) {
-        transactionTableController.unmount()
-      }
+      transactionTableController.unmount()
     })
     watch(accountId, () => {
       if (mounted) {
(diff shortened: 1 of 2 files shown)


src/pages/AccountDetails.vue
ericleponner, line 1: That means:
1) conditional TableController.mount()
2) inconditional TableController.unmount()
Mmm… this is asymmetric … not sure it's the right way to fix this.

Sheng-Long: There are many existing checks in all of the unmount lifecycles for TableController so that unmounting can be done at any time, seemingly

ericleponner: Ok, I now get a better understanding of the issue.

`onBeforeUnmount()` callback assumes that if `accountId.value` is null, this has been detected by `watch()` callback and `transactionTableController.unmmount()` has already been called
But, for some reason, this assumption is false: `accountId.value `is reset to null but `watch()` callback is not invoked at all (normal behavior of vuejs unmounting process?).

So, as you suggested it, test `if (accountId.value !== null)` is wrong.

In place of unconditional `transactionTableController.unmount()` and the patch in `TableController`, I would prefer the following and keep `TableController` unchanged:
```
onBeforeUnmount(() => {
   mounted = false
   if (transactionTableController.mounted.value) {
     transactionTableController.unmount()
   }
}

```
This preserves a design invariant in `TableController`: `refreshBuffer()` should be invoked only when controller is mounted.

PR #775, opened 2023-11-28
00694 display both raw and b64-decoded values of memo when relevant

### LINKED ISSUE #694: Incorrect memo display.
### Description

Got some small visual bug on hashscan. Account memo specificly for phrase "adap" displayed as strange symbol. Worth to mention that it's only word that cause this bug for me. Any other word works well.
Test account with reproduced issue: https://hashscan.io/testnet/account/0.0.1709113
Also screenshot attached.
image

### Steps to reproduce

1. Create new account with `.setAccountMemo('adap')`
2. Check visual of memo on Hashscan.

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

latest

### Operating system

None

svienot:
**Description**:

In AccountDetails and ContractDetails views, when we find that the memo  is candidate to Base64 decoding, we now show both the raw value and the decoded value. So if the value is, in fact, not encoded, it is still legible.

This happens only in case the value has not been interpreted (as a URL, an IPFS hash...) since that would mean we were indeed not confused about the encoding.

**Related issue(s)**:

Fixes #694 

**Notes for reviewer**:

Visible on staging server.
Examples:
 - actual B64 encoding Screenshot 2023-11-28 at 12 42 37
 - not a B64 encoding Screenshot 2023-11-28 at 12 44 00

**Checklist**

### DIFF
--- a/src/components/values/BlobValue.vue
+++ b/src/components/values/BlobValue.vue
@@ -37,11 +37,13 @@
       <span id="blob-main">
         {{ (b64EncodingFound && showBase64AsExtra) ? blobValue : decodedValue }}
       </span>
-      <div v-if="b64EncodingFound && showBase64AsExtra"  id="blob-extra" class="h-is-extra-text h-is-text-size-3 mt-1">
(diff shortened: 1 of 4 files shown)


ericleponner (approved):
LG

PR #785, opened 2023-11-30
00784 reactivity problem with the contract verification status

### LINKED ISSUE #784: Reactivity problem with the contract verification status
### Description

A contract may appear as UNVERIFIED when the ContractDetails page is (re)loaded, although its status is VERIFIED. This never happens when navigating from another page. This is very sensitive to page initialisation dynamics (may vary with browser, etc...)

### Steps to reproduce

Search contract 0.0.5840019 on testnet
Contract details page shows it as verified
Re-load the page (possibly several times)
The contract appears as not verified

### Additional context

Easy to see on Chrome. Not so much on Safari.

### Hedera network

mainnet, testnet, previewnet

### Version

v23.11.0-10-g4044f6c 

### Operating system

None

svienot:
**Description**:

Adjust the reactivity of the ContractDetails view such that:
 - the (re)loading issue in 784 is fixed
 - we avoid the flashing effect: displaying NOT-VERIFIED when page is loaded and VERIFIED once the verification status has been obtained from sourcify.

**Related issue(s)**:

Fixes #784 

**Checklist**

### DIFF
--- a/src/components/contract/ContractByteCodeSection.vue
+++ b/src/components/contract/ContractByteCodeSection.vue
@@ -28,15 +28,15 @@
         <template v-slot:title>
             <div class="is-flex is-align-items-center is-flex-wrap-wrap">
                 <span class="h-is-secondary-title mr-3">Contract Bytecode</span>
-                <div v-if="isVerificationEnabled" class="h-is-text-size-2 mt-1">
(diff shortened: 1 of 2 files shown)


ericleponner (approved):
Looks good.

PR #796, opened 2023-12-11
chore: add configuration and run target for running on HTTP.

svienot:
**Description**:

Add HTTP set-up which is needed to be able to run/debug explorer and sourcify locally.


ericleponner (approved):
Looks good.

PR #803, opened 2023-12-14
fix: miscellaneous fixes

svienot:
**Description**:

A couple unrelated miscellaneous fixes:
 - always show `REFRESH PAUSED` when tables are not in play mode
 - always sort transactions by `Nonce` in TransactionById view
 - display `Nonce` or `Relationship` column only when relevant in TransactionById view
 - get rid of code related to IPFS/swarm since we do not leverage these for now

**Notes for reviewer**:

Deployed to staging server


src/components/PlayPauseButton.vue
ericleponner, line 1: `isAutoStopped` computed could be removed I guess

ericleponner (approved):

src/components/PlayPauseButton.vue
svienot: Indeed! I have removed that computed.

Issue #756, opened 2023-11-15
Smart contract verification phase 2 does not accept metadata file generated by Foundry-rs


quiet-node:
### Description

Seems like metadata generated by foundry-rs by is structured differently than other tools like Solc, Hardhat, etc. so the FileImporter doesn't accept the metadata from foundry. 

Here is an example of metadata file generated by foundry
[Counter.json](https://github.com/hashgraph/hedera-mirror-node-explorer/files/13370776/Counter.json)

The contract: 
``` solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

contract Counter {
    uint256 public number;

    function setNumber(uint256 newNumber) public {
        number = newNumber;
    }

    function increment() public {
        number++;
    }
}
```

And here is the link to the hashscan page of the contract:  https://hashscan-latest.hedera-devops.com/testnet/contract/0.0.5893811

### Steps to reproduce

1. Deploy a contract by foundry-rs
2. Go to the hashscan page of the contract -> click verify contract -> drop the metadata file generated by foundry to the file. chooser -> nothing is showing

### Hedera network

testnet

### Version

staging server

### Operating system

macOS

ericleponner:
Could you add `Counter.sol` to the issue ?

quiet-node:
@ericleponner I have updated the issue to include the contract. The issue does not accept .sol file so I just pasted the raw code there.

svienot:
Will be fixed once #799 is integrated.
Fix available on staging server.

ericleponner:
This issue is fixed by #811.

PR #811, opened 2023-12-19
feat: Contract Verification dialog should make use of Sourcify Session Verification API

### LINKED ISSUE #799: Contract Verification dialog should make use of Sourcify Session Verification API
### Problem

Currently `Contract Verification` dialog performs verification flow by:
- using some smartness on client side (by running `soljs` compiler)
- calling Sourcify [Stateless Verification](https://server-verify.hashscan.io/api-docs/#/Stateless%20Verification) API

Smartness on client side tries to pre-verify contract and determine target contract name.

Benefit of this approach is that user is aware before clicking on `VERIFY CONTRACT` button (which is a one-way operation). Drawback is that client side smartness does not cover all verification cases that are supported by Sourcify server. Maintaining parity between the two will be substantial effort.

Proposed alternative is to:
- remove client side smartness from `Contract Verification` dialog
- make use of Sourcify [Session Verification](https://server-verify.hashscan.io/api-docs/#/Session%20Verification) API and its new `dryrun` parameter (see [#90](https://github.com/hashgraph/hedera-sourcify/issues/90))

### Solution

Use Sourcify [Session Verification](https://server-verify.hashscan.io/api-docs/#/Session%20Verification) API and its `dryrun` parameter.

### Alternatives

_No response_

ericleponner:
**Description**:

With the changes below, Explorer now uses on Sourcify Session Verification API.

In Explorer, `Contract Verification` dialog performs a  verification preview which enables user to known whether verification will be a `full` or `partial` *before* publishing contract on Hedera Sourcify service.

So far, this preview was an approximation of the verification process performed by Hedera Sourcify Server. By using Sourcify Session Verification (an `dryrun` flag), Explorer preview follow exactly the same process as Hedera Sourcify Server.

**Related issue(s)**:

Fixes #799 , #756

**Checklist**


svienot (approved):
Looks good.

PR #824, opened 2023-12-28
feat: #822 collapsible sections

### LINKED ISSUE #822: Collapsible sections in details views.
Implement collapsible sections per design spec:
Screenshot 2023-12-27 at 19 30 55

We'll start with the following:
- we make collapsible all the sections on all the pages but the Dashboard page and the other main pages containing a single table (i.e. Transactions, Accounts...)
- sections are not collapsed initially
- user preference for each section is saved in local storage

We might later decide that by default certain sections are collapsed by default (e.g. because considered only for advanced users, or for a specific persona).

svienot:
**Description**:

- Make sections collapsible per spec. This is done for all pages but main pages which contain a single table (and the Dashboard page)
- Remove navigation links from Staking page

**Related issue(s)**:

Fixes #822

**Notes for reviewer**:
- Deployed on staging server
- Sample screenshot:  Screenshot 2023-12-28 at 11 46 55

### DIFF
--- a/src/AppStorage.ts
+++ b/src/AppStorage.ts
@@ -119,6 +119,21 @@ export class AppStorage {
     }
 
     //
+    // sections collapsed state
+    //
+
+    private static readonly COLLAPSED_STATE_KEY = 'collapsed'
+
+    public static getCollapsedState(section: string): boolean | null {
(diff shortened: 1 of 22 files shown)


ericleponner (approved):
Looks good.

PR #833, opened 2024-01-05
fix: error decoding not working as expected on verified contract

### LINKED ISSUE #829: Error decoding not working as expected on verified contract
### Description

For verified contracts, it appears that require statements that revert with a string are not decoded properly in the hashscan interface. We expect that a smart contract reverting on a solidity require statement: require(1 ==0, 'abcd') should render 'abcd' in hashscan as the error reason

### Steps to reproduce

Call a verified smart contract that reverts using a require statement.

### Additional context

An example of a transaction we suspect should revert with a string:
https://hashscan.io/mainnet/transaction/1704390315.881373003

Line where suspected revert occurs in the smart contract:
https://github.com/saucerswaplabs/saucerswap-periphery/blob/1f0e50ef94dee14dc616bfe4bf5315029f21c18e/contracts/UniswapV2Router02.sol#L48

Another example on testnet where a smart contract calls another smart contract, resulting in a revert (note that the string is decoded in the call trace on page 3):
https://hashscan.io/testnet/transaction/1704318176.839930003

### Hedera network

mainnet, testnet

### Version

latest

### Operating system

macOS

ericleponner:
**Description**:

Changes below fix #829: `FunctionError` vue now displays error arguments.
They also include some new unit test cases in `FunctionCallAnalyzer.spec.ts`.

**Related issue(s)**:

Fixes #829 

**Notes for reviewer**:
See #829 for testing data.

**Checklist**


svienot (approved):
Looks good to me.

Issue #804, opened 2023-12-15
Raw data or links to mirror node api on transactions and/or all entities


Sheng-Long:
### Problem

As a developer, sometimes I'd rather look at the raw data for transactions and/or entities that is provided by the api rather than the formatted data provided by Hashscan.

### Solution

Either:
- Provide collapsible raw data for transactions and/or entities at the bottom of the page
or
- Provide a link to the mirror node api for the specified transaction and/or entity

### Alternatives

The only current alternative is to navigate to the correct page on the mirror node api

PR #875, opened 2024-01-30
feat: add link to mirror node data for entities

Sheng-Long:
**Description**:
Add link to mirror node data for entities

**Related issue(s)**: #804 

Fixes #

**Notes for reviewer**:

**Checklist**


Sheng-Long:
Not entirely sure what's going on in the unit tests, will check it out

src/components/MirrorLink.vue
```diff
@@ -0,0 +1,53 @@
...
+<!-- --------------------------------------------------------------------------------------------------------------- -->
+<!--                                                     TEMPLATE                                                    -->
+<!-- --------------------------------------------------------------------------------------------------------------- -->
+
+<template>
+  <a :href="`${networkUrl}api/v1/${entityUrl}/${loc}`">
+    <span class="is-numeric">Mirror Data</span>
+  </a>
+</template>
+
```
svienot, line 30: I suggest to use the following to make the link less prominent (it is for the advanced user) and also not force the user to navigate back to the explorer page:

```
  <a :href="`${networkUrl}api/v1/${entityUrl}/${loc}`" target="_blank">
      <p class="has-text-right has-text-grey h-is-property-text">Show raw data</p>
  </a>
```

src/components/MirrorLink.vue
```diff
@@ -0,0 +1,53 @@
...
+  name: "MirrorLink",
+  components: {},
+  props: {
+    network: String,
+    entityUrl: String,
+    loc: String,
+  },
+
+  setup(props, context) {
+    const networkUrl = props.network == "mainnet" ?  "https://mainnet-public.mirrornode.hedera.com/" : "https://testnet.mirrornode.hedera.com/"
```
svienot, line 48: We need to take care of previewnet as well.

src/components/contract/ContractResult.vue
```diff
@@ -118,6 +118,9 @@
 
     <ContractResultLogs :logs="contractResult?.logs" :block-number="blockNumber" :transaction-hash="transactionHash"/>
 
+    <!--Workaround to ensure margin bottom-->
+    <div />
+
```
svienot, line 123: @Sheng-Long  you may rebase your branch onto main. You will get a small fix which solves such margin discrepancies at the end of the pages.
You can get rid of that workaround `<div>`

src/pages/TopicDetails.vue
```diff
@@ -45,6 +45,10 @@
 
     </DashboardCard>
 
+    <div style="height: 24px" />
```
svienot, line 48: You can also get rid of that `<div>` element -- see previous comment.

svienot (changes requested):
See my couple comments.
Other than that, looks good. This is going to be very useful (for us, to begin with)!

svienot (approved):
LG

PR #880, opened 2024-01-30
feat: Explorer could interpret Solidity method selectors

### LINKED ISSUE #696: Explorer could interpret Solidity method selectors - giving method signature for method calls, other situations
### Problem

Explorer could provide more information for contract calls/returns/results by interpreting fields.  For example, the first 4 bytes of a contract call's calldata is most often a 4-byte encoded method selector.  And it would be great to know what method was being called.

For example, picked at random just now, consider [this contract call](https://hashscan.io/mainnet/transaction/1695254373.390339008?p2=1&k2=1695254445.984780953&p1=1&k1=1695254445.957518951):

image

That leading `618d c65e` in the "input - function & args" hex string is actually a call to `redirectForToken(address,bytes)` - in other words, our HTS token proxy contract.  (Not too interesting in this case, but I _did_ say I picked it at random).  I found that out by [looking here](https://www.4byte.directory/signatures/?bytes4_signature=618dc65e):

image

It would be nice if the explorer would interpret method selectors for me when it could.  Also, events.

### Solution

For various contract byte buffers (hexified) displayed by the Explorer it could find method signatures (or event signatures) in "likely places" (like the first 4 bytes of a calldata) and provide the method name/signature by looking it up in a database.

For example:

`https://4byte.directory` is a public (and open-source) database of Ethereum method selectors (and event selectors).  ([Github repo here](https://github.com/gobitfly/ethereum-function-signature-registry)).  It can be updated manually by a contract's author, or I think Sourcify will do it for you.  Anyway, there are currently ~1.2M signatures in its database.

* And it's not the only Ethereum method selector database out there.  There's also [etherface](https://github.com/volsa/etherface) and others. But considering here only 4byte:

The database itself is unfortunately not directly available.  But given the 4byte API it is easily scrapable.  And given an initial scrape, it can be incrementally scraped for new entries on a schedule (e.g., nightly). (This is explicitly allowed by its license and terms of service, as seen at the bottom of its home page.)

Event signatures are uploaded there too, which would be useful in display Ethereum event logs.

### Alternatives

_No response_

ericleponner:
**Description**:

With changes below, Explorer now queries [4byte.directory](https://www.4byte.directory) and tries to decode call input/output. `Contract Result` and `Call Trace` sections in `Transaction Details` are now able to display decoded input and output even if corresponding contract is unverified.

Changes also include additional unit tests for `FunctionCallAnalyzer` class.

Note: Event decoding is not handled by these changes ; this will be done in a second phase.

**Related issue(s)**:

Fixes #696

**Notes for reviewer**:

This [transaction](https://hashscan-latest.hedera-devops.com/mainnet/transaction/1706289064.687696003) gives a pretty good view of the feature (see `Call Trace` section).

**Checklist**


svienot (approved):
Looks good to me!

PR #883, opened 2024-01-31
chore: tentative fix for random break in unit tests

ericleponner:
**Description**:

Changes below fix `MetadataCache.clear()` : this is suspected to be root cause of random breaks in unit tests.

**Checklist**


svienot (approved):
LGTM

PR #911, opened 2024-03-01
chore: added Dialog component and associated classes (for further use)

ericleponner:
**Description**:

Changes below add some new classes for managing modal dialog.
This new code will be used to re-factor all our existing dialogs.


svienot (approved):
Looks good.

PR #917, opened 2024-03-06
chore: improve Tabs component

svienot:
**Description**:

Simple changes to identify tabs by a string instead of their index, and guard against wrong value obtained from user prefs.

**Checklist**


ericleponner (approved):
Looks good.

PR #923, opened 2024-03-08
chore: remove use of 'app' query parameter in AccountDetails

svienot:
**Description**:

Remove dead code (this was once used to open the ApproveAllowanceDialog directly from the Staking page).


ericleponner (approved):
Looks good.

PR #935, opened 2024-03-21
fix: prevent editing allowance when token is not associated

### LINKED ISSUE #584: Allow for changing allowance for token that is no longer associated
### Description

As a user if I set an allowance for a token that I was once associated with but no longer am I would like to still manage that allowance on hashscan to for example delete it.

Look at this [account's allowances ](https://hashscan.io/mainnet/account/0.0.1055570), it is no longer associated with token [0.0.2178426](https://hashscan.io/mainnet/token/0.0.2178426), however the allowance that was previously given to contract 0.0.2178425 still shows under it's allowances. Users may want to delete such allowances.

### Steps to reproduce

1. Associate with a token
2. Grant an allowance
3. Dissociate from that same token
4. See that the allowance still shows on Hashscan, however Hashscan does not allow the user to edit/change the allowance since they're no longer associated with the token.

### Additional context

_No response_

### Hedera network

mainnet

### Version

Hashscan  Built Fri, 05 May 2023 08:06:30 GMT

### Operating system

None

svienot:
**Description**:

- Display (i) icon and tooltip instead of edit icon in the `Token Allowances` table when the token is not associated (any more) with the account.
- Fix display of token amount in the `ApproveAllowanceDialog` when the token has decimals.
- Display token symbol in dialog confirmation messages
- Clarify confirmation messages when the edit is in fact a suppression (amount set to 0)

**Related issue(s)**:

Fixes #584

**Notes for reviewer**:

See screen captures in associated issue #584

### DIFF
--- a/src/components/InfoTooltip.vue
+++ b/src/components/InfoTooltip.vue
@@ -27,8 +27,8 @@
              :delay="delay"
              multiline="multiline"
              :position="position ?? 'auto'"
-             class="h-tooltip ml-1">
-    <span class="icon is-small h-is-extra-text"><i class="fas fa-info-circle"></i></span>
+             class="h-tooltip">
+    <span class="icon is-small h-is-property-text h-is-extra-text"><i class="fas fa-info-circle"></i></span>
   </o-tooltip>
 </template>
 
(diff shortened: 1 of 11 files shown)


ericleponner (approved):
Looks good

PR #936, opened 2024-03-21
feat: action for downloading ABI of a verified contract

### LINKED ISSUE #934: Provide easy access to ABIs for verified contracts
### Problem

At SaucerSwap, we help people with integration of our smart contracts into their platforms and apps.  We recently noticed that the ABIs for verified contracts were not easily accessible on Hashscan.  It looks like they are still available via downloading the entire source but it's not immediately obvious you can get the metadata.json that way.  When people want to interact with contracts via ethers.js or something similar, the ABI as a file is the easiest way to go.  If they are not smart contract developers, to ask them to checkout and compile our repos is outside of their normal work flow.

### Solution

It would be great if the ABI was more easily accessible to users.  It would also be great if the contract ABI was formatted in the the same way the the compiler outputs them vs metadata.json.   In tutorials, they typically refer to the compiler output when they interact with the deployed smart contract.  I believe this will reduce friction for people that may not be familiar with smart contract development.

### Alternatives

We considered asking them to build the project, but some developers are not familiar with smart contracts and may have difficulty.  This is another point of friction for us and to ask them makes the process more cumbersome.

We considered hosting the ABIs in GitHub but Hashscan should be the source of truth for them since it verifies contracts.

Screenshot from 2024-03-20 12-34-36

https://www.web3.university/tracks/create-a-smart-contract/interact-with-your-smart-contract

ericleponner:
**Description**:
Changes below add an action for downloading ABI to `Contract Details` page (in ABI tab).
This button is currently visible on `previewnet` only.
It will be enabled on `testnet` and `mainnet` later.

**Related issue(s)**:

Fixes #934 

**Notes for reviewer**:

image

### DIFF
--- a/src/components/contract/ContractByteCodeSection.vue
+++ b/src/components/contract/ContractByteCodeSection.vue
@@ -103,7 +103,7 @@
                         <input type="checkbox" v-model="showHexaOpcode">
                     </label>
                 </div>
-                <div v-else-if="selectedOption==='abi'" class="is-flex is-justify-content-end"><DownloadButton @click="handleDownloadABI"/></div>
+                <div v-else-if="selectedOption==='abi' && showDownloadABI" class="is-flex is-justify-content-end"><DownloadButton @click="handleDownloadABI"/></div>
             </div>
             <SourceCodeValue  v-if="isVerified && selectedOption==='source'" class="mt-3"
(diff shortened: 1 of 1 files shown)


svienot (approved):
Looks good !

PR #965, opened 2024-04-09
chore: improve source formatting

svienot:
**Description**:

Applies the default IntelliJ code formatting to source files (including tests). A few files need to be manually edited to avoid breaking tests sensitive to extra spaces sometimes introduced by reformatting (with string constants embedded in div).

The second commit is where manual changes are applied. The first and third commits are the result of pure IDE reformatting.


ericleponner (approved):
LGTM

PR #978, opened 2024-04-18
feat: proxy contract support in ABI tab

### LINKED ISSUE #964: Enable to call implementation contract from a proxy contract
### Problem

This is a follow-up of this [request](https://github.com/hashgraph/hedera-mirror-node-explorer/issues/782#issuecomment-1908955006).

### Solution

There are some ways to:
1) detect that a contract is a proxy 
2) fetch id / address of the implementation contract from proxy bytecode
but these are both heuristic and need to be investigated.

May be a first approach could be to rely on user's knowledge:
user could be asked to provide id/address of the implementation contract before calling implementation methods from the proxy. Feedback is welcome.

### Alternatives

_No response_

ericleponner:
**Description**:

Changes below add support for proxy contract to the `ABI` tab (in `Contract Details` page) and matching unit tests.

Explorer now reads contract state (as defined in [IEP 1967](https://eips.ethereum.org/EIPS/eip-1967)) to determine if a contract is a proxy or not.
When contract is a proxy:
- `Contract Bytecode` section displays two new fields:
    - `Proxying to Logic Contract` which displays logic contract id
    - `Proxying with Admin Contract` which displays admin contract id
- `ABI` tab displays a new switch named `Show logic contract ABI`
- when switch is on:
   - ABI of the logic contract is displayed in place of the proxy ABI
   - functions (from logic ABI) are called through the proxy contract

image

**Related issue(s)**:

Fixes #964

**Notes for reviewer**:

Sample proxy contract on `mainnet`: [0.0.4544494](https://hashscan.io/mainnet/contract/0.0.4544494)

**Checklist**


svienot (approved):
Looks good !

PR #989, opened 2024-04-24
fix: decoded message using ethers.toUtf8String()

### LINKED ISSUE #904: Enhancing Error Message Readability on Hashscan for Reverted Smart Contract Calls
### Problem

**Description:**
Currently, when a smart contract call is reverted, the error message displayed on Hashscan is well-readable if the error message is either `Error(string)` or `Panic(uint256)`. However, for custom errors, the display lacks human readability, as it doesn't decode the custom data into a readable format.

**Example:**
```solidity
function buy(uint amount) public payable {
    if (amount > msg.value / 2 ether)
        revert("Not enough Ether provided.");
}
```
In this example, the error generates an `Error(string)` data, which Hashscan displays nicely.
image

**link to example transaction:** https://hashscan.io/testnet/transaction/1708618373.183662003

**Proposal:**
Enhance the error message display for custom errors such as:
```solidity
error Unauthorized(address);
function withdraw() public {
    if (msg.sender != owner)
        revert Unauthorized(msg.sender);
}
```
Currently, Hashscan decodes custom error data into a non-human-readable format.
image

**link to example transaction:** https://hashscan.io/testnet/transaction/1708620069.753947441

This issue proposes improving this to make custom error messages more easily understandable to developers.

**Expected Outcome:**
With this enhancement, developers will benefit from better readability of error messages, especially for custom errors, leading to improved debugging and development experiences when interacting with smart contracts on Hashscan.

### Solution

- If the contract is not verified and the error is caused by a custom error, display the data hash
~~- If the contract is verified, decode the error using its ABI~~

### Alternatives

_No response_

quiet-node:
**Description**:
This PR adds an extra decoding logic to decode error message using ethers.toUtf8String(). 

**UI showcase:
for example, with this transaction on previewnet https://hashscan.io/previewnet/transaction/1713967052.604875873, the error message only shows the hexa data like below
image

but using the extra logic mentioned above, it can be decoded like below
image

**Related issue(s)**:

Fixes #904

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/schemas/HederaUtils.ts
+++ b/src/schemas/HederaUtils.ts
@@ -158,7 +158,7 @@ export function decodeSolidityErrorMessage(message: string | null): string | nul
         } else if (!message.startsWith("0x")) {
             result = message
         } else {
-            result = null;
+            result = ethers.toUtf8String(message);
         }
     } catch (reason) {
         result = null


src/schemas/HederaUtils.ts
```diff
@@ -158,7 +158,7 @@ export function decodeSolidityErrorMessage(message: string | null): string | nul
         } else if (!message.startsWith("0x")) {
             result = message
         } else {
-            result = null;
+            result = ethers.toUtf8String(message);
```
Nana-EC, line 161: This is good.

Does decoding always result in legible text?
If not should we make this a flow where the user could click a button to decode on desire?
The hex might still be valuable to some devs

Nana-EC (commented):
Nice simple fix.
One consideration that might make it more complex 😆

src/schemas/HederaUtils.ts
quiet-node: Right so ethers.toUtf8String will either return an legible text or it will throw a decoding exception, and this is wrapped in a try catch, so when the program catch the exception, the `result` will be set to `null`, and when `result = null`, hashscan will show the original hex string of the error message.

ericleponner: Looks good to me.

ericleponner (approved):

ericleponner (changes requested):
After investigation, I think this fix is not needed.
In fact, flow of control should not land in this method when it's a call to HTS.
Root cause is elsewhere.

ericleponner:
The way `Call Trace` section performs output and error decoding is deficient.
I changed `Call Trace` section to use the same logic as in `Contract Result` section.
See #999 and #1000 for details.

Fallback proposed in this PR is no longer needed.
I even think we should not do it (we would decode bytes that are not UTF8).
Closing the PR.

PR #993, opened 2024-04-26
fix: added a new decoding logic for redirectForToken() (#921)

### LINKED ISSUE #921: HTS child transactions throw out-of-bound exception when decoding function parameters for `redirectForToken()`
### Description

The function parameters returned from `redirectForToken()` are different from the other functions in the HTS contract. While other functions encode the function parameters and pad zeros to the hex string to make each slot 32 bytes, encoded function parameters of the redirectForToken() only contains raw data. For instance, consider the this encoded function parameters below

`0x618dc65e0000000000000000000000000000000000376d7c06fdde03`

The first 4 bytes represent the function selector of `redirectForToken()` , which can be ignored for now. The next 20 bytes represent the `tokenAddress`, while the last 4 bytes signify the `ERC20.name()` function. Thus, the total size of the parameters is 24 bytes, which is less than the minimum 32 bytes.

This discrepancy leads to an out-of-bound exception when trying to decode it using the [ethers.AbiCoder.defaultAbiCoder.decode()](https://docs.ethers.org/v6/api/abi/abi-coder/#AbiCoder-decode) method from [ethersjs](https://docs.ethers.org/v6/).

### Steps to reproduce

1. Make a transaction by calling `redirectForToken()` with any tokenAddress and encodedFunctionSelector data
2. Go the the child transaction of the transaction
3. Observe the `out-of-bound` exception thrown while attempting to decode encoded function parameters

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet, other

### Version

latest

### Operating system

None

### LINKED ISSUE #984: Proper decoding of the Precompile contract inputs
### Problem

From bugbytes

Discord: https://discord.com/channels/373889138199494658/616725704284962837/1230862128911618048

Comment: one really nice thing that could be improved in hashscan is proper decoding of the Precompile contract inputs?  Particularly 0x0167, it would be very useful for debugging

image12

### Solution

TBD

### Alternatives

_No response_

quiet-node:
**Description**:
This PR introduces new decoding logic for parsing function inputs and outputs of the `redirectForToken()` method from the proxy contract deployed at address 0x167 within the HTS system.

The enhanced logic is designed to discern transaction results and determine the validity of a `redirectForToken()` transaction. This involves verifying that the transaction is directed at the HTS system contract(`0x167`) and that its sighash matches redirectForToken()'s sighash, `0x618dc65e`.

Regarding transaction inputs, the `redirectForToken()` function's input arguments (inputArgs) consist solely of raw data without padding bytes. Consequently, `ethers.AbiCoder.defaultAbiCoder().decode()` cannot function as expected and throws a `Decoding Error (data out-of-bound)`. According to the ABI, the inputArgs comprise two components: the first 20 bytes denote the `tokenAddress`, while the remainder represents the `encodedFunctionSelector`. The new decoding logic capitalizes on this structure by segregating the two components from the inputArgs. This facilitates the assembly of transaction inputs into an `ethers.Result` object encompassing both components, thus ensuring that zero padding is applied to each slot. Consequently, `ethers.AbiCoder.defaultAbiCoder().decode()` can successfully decode function inputs.

In terms of transaction outputs, the outputs returned by the proxy contract's `redirectForToken()` method correspond to the standards set forth by ERC20 and ERC721 tokens. The updated decoder compiles a list of supported ERC20 and ERC721 standard methods (as delineated in [HIP-218](https://hips.hedera.com/hip/hip-218)) to decode the outputs of the `redirectForToken()` function.

### Before the fix:
image

image

### After the fix:
image

image

**Related issue(s)**:

Fixes #921
Fixes #984 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/schemas/HederaUtils.ts
+++ b/src/schemas/HederaUtils.ts
@@ -252,25 +252,19 @@ export function lookupNFTTransfer(transaction: Transaction, tokenId: string): Nf
     return result
 }
 
-export function decodeRedirectForTokenInput(inputArgs: string): ethers.Result {
-    const tokenAddress = `0x${inputArgs.slice(2, 42)}`
-    const encodedFunctionSelector = `0x${inputArgs.slice(42)}`
-    return new ethers.Result(tokenAddress, encodedFunctionSelector)
+export function decodeRedirectForTokenInput(functionFragment: ethers.FunctionFragment, inputArgs: string): ethers.Result {
+    try {
+        return ethers.AbiCoder.defaultAbiCoder().decode(functionFragment.inputs, inputArgs)
+    } catch (e) {
+        const tokenAddress = `0x${inputArgs.slice(2, 42)}`
+        const encodedFunctionSelector = `0x${inputArgs.slice(42)}`
+        return new ethers.Result(tokenAddress, encodedFunctionSelector)
+    }
 }
 
 export function resolveFunctionFragmentForHTSProxyContract(functionFragment: ethers.FunctionFragment, inputArgs: string): ethers.FunctionFragment {
-    let encodedFunction4BytesSignature = ""
-
-    try {
-        const inputResult = ethers.AbiCoder.defaultAbiCoder().decode(functionFragment.inputs, inputArgs)
-        encodedFunction4BytesSignature = inputResult[1].slice(0, 10)
-    } catch (failure) {
-        const f = failure as ethers.EthersError
-        if (f.code === "BUFFER_OVERRUN" || f.code === "INVALID_ARGUMENT") {
-            const inputResult = decodeRedirectForTokenInput(inputArgs)
-            encodedFunction4BytesSignature = inputResult[1].slice(0, 10)
-        }
-    }
+    const inputResult = decodeRedirectForTokenInput(functionFragment, inputArgs)
+    const encodedFunction4BytesSignature = inputResult[1].slice(0, 10)
 
     // @notice this is the list of supported method which is collected based on HIP-218 https://hips.hedera.com/hip/hip-218
     const ABI_FOR_SUPPORTED_METHODS = [
@@ -300,7 +294,7 @@ export function resolveFunctionFragmentForHTSProxyContract(functionFragment: eth
     return ethers.FunctionFragment.from({...functionFragment, outputs: iface.getFunction(encodedFunction4BytesSignature)?.outputs})
 }
 
-export function isRedirectForTokenTx(contractId: string|null, functionHash: string|null): boolean {
+export function isRedirectForTokenTx(contractId: string, functionHash: string): boolean {
     return contractId === HTS_PRECOMPILE_CONTRACT_ID &&
         functionHash === REDIRECT_FOR_TOKEN_FUNCTION_SIGHASH
 }
--- a/src/utils/analyzer/FunctionCallAnalyzer.ts
+++ b/src/utils/analyzer/FunctionCallAnalyzer.ts
@@ -302,21 +302,18 @@ export class FunctionCallAnalyzer {
 
         if (ff !== null && inputArgs !== null) {
             try {
-                this.inputResult.value = ethers.AbiCoder.defaultAbiCoder().decode(ff.inputs, inputArgs)
-                this.inputDecodingFailure.value = null
-            } catch(failure) {
-                const f = failure as ethers.EthersError
-                // please refer to the ticket below for more information on this logic for redirectForToken(address,bytes) method on HTS System Contract 
-                // https://github.com/hashgraph/hedera-mirror-node-explorer/issues/921
-                if (
-                    (f.code === "BUFFER_OVERRUN" || f.code === "INVALID_ARGUMENT") &&
-                    isRedirectForTokenTx(contractId, functionHash)
-                ) {
-                    this.inputResult.value = decodeRedirectForTokenInput(inputArgs)
-                } else {
-                    this.inputResult.value = null
-                    this.inputDecodingFailure.value = failure
+                if (contractId !== null && functionHash !== null) {
+                    if (isRedirectForTokenTx(contractId, functionHash)) {
+                        this.inputResult.value = decodeRedirectForTokenInput(ff, inputArgs)
+                        this.inputDecodingFailure.value = null
+                    } else {
+                        this.inputResult.value = ethers.AbiCoder.defaultAbiCoder().decode(ff.inputs, inputArgs)
+                        this.inputDecodingFailure.value = null
+                    }
                 }
+            } catch (failure) {
+                this.inputResult.value = null
+                this.inputDecodingFailure.value = failure
             }
         } else {
             this.inputResult.value = null


src/schemas/HederaUtils.ts
```diff
@@ -249,3 +249,51 @@ export function lookupNFTTransfer(transaction: Transaction, tokenId: string): Nf
...
+        "function allowance(address owner, address spender) external view returns (uint256 allowance)",
+        "function safeTransferFrom(address _from, address _to, uint256 _tokenId, bytes data) external payable",
+        "function isApprovedForAll(address _owner, address _operator) external view returns (bool isApprovedForAll)",
+        "function transferFrom(address sender, address recipient, uint256 amount) external returns (bool transferFrom)",
+        "function tokenOfOwnerByIndex(address _owner, uint256 _index) external view returns (uint256 tokenOfOwnerByIndex)",
+      ];
+      
+    const iface = new ethers.Interface(ABI_FOR_SUPPORTED_METHODS)
+    return ethers.FunctionFragment.from({...functionFragment, outputs: iface.getFunction(encodedFunction4BytesSignature)?.outputs})
+}
```
ericleponner, line 299: We could encapsulate `redirectForToken()` detection in a method and use it (two times) in `FunctionCallAnalyzer`:

```
export function isRedirectToken(contractId: string, functionHash: string): boolean {
    return contractId === HTS_PRECOMPILE_CONTRACT_ID &&
        functionHash === REDIRECT_FOR_TOKEN_FUNCTION_SIGHASH
}

```

src/utils/analyzer/FunctionCallAnalyzer.ts
```diff
@@ -236,13 +237,30 @@ export class FunctionCallAnalyzer {
...
+                    if (
+                        currentFunctionFragment !== null &&
+                        contractId === HTS_PRECOMPILE_CONTRACT_ID &&
+                        functionHash === REDIRECT_FOR_TOKEN_FUNCTION_SIGHASH
+                    ) {
+                        currentFunctionFragment = resolveFunctionFragmentForHTSProxyContract(currentFunctionFragment, inputArgs)
+                    }
+
+                    this.functionFragment.value = currentFunctionFragment
+                } catch(failure) {
```
ericleponner, line 263: Here we could use isRedirectToken() method and update control flow more explicitly:

```
         if (functionHash !== null && inputArgs !== null && contractId !== null) {
            if (i !== null) {
                try {
                    const ff = i.getFunction(functionHash)
                    if (isRedirectToken(contractId, functionHash) && ff !== null) {
                        this.functionFragment.value = resolveFunctionFragmentForHTSProxyContract(ff, inputArgs)
                        this.functionDecodingFailure.value = null
                        this.is4byteFunctionFragment.value = false
                    } else {
                        this.functionFragment.value = ff
                        this.functionDecodingFailure.value = null
                        this.is4byteFunctionFragment.value = false
                    }
                } catch(failure) {
                    this.functionFragment.value = null
                    this.functionDecodingFailure.value = failure
                    this.is4byteFunctionFragment.value = false
                }
            } else if (r !== null && r.results.length >= 1) {

```

src/utils/analyzer/FunctionCallAnalyzer.ts
```diff
@@ -284,13 +302,27 @@ export class FunctionCallAnalyzer {
...
+                if (
+                    (f.code === "BUFFER_OVERRUN" || f.code === "INVALID_ARGUMENT") &&
+                    contractId === HTS_PRECOMPILE_CONTRACT_ID &&
+                    functionHash === REDIRECT_FOR_TOKEN_FUNCTION_SIGHASH
+                ) {
+                    this.inputResult.value = decodeRedirectForTokenInput(inputArgs)
+                } else {
+                    this.inputResult.value = null
+                    this.inputDecodingFailure.value = failure
+                }
```
ericleponner, line 325: Here too we can re-use `isRedirectForToken()` and have a straight if-then-else flow:

```
            try {
                if (contractId !== null && functionHash !==  null) {
                    if (isRedirectToken(contractId, functionHash)) {
                        this.inputResult.value = decodeRedirectForTokenInput(inputArgs)
                        this.inputDecodingFailure.value = null
                    } else {
                        this.inputResult.value = ethers.AbiCoder.defaultAbiCoder().decode(ff.inputs, inputArgs)
                        this.inputDecodingFailure.value = null
                    }
                } else {
                    this.inputResult.value = null
                    this.inputDecodingFailure.value = null
                }
            } catch(failure) {
                this.inputResult.value = null
                this.inputDecodingFailure.value = failure
            }

```

src/schemas/HederaUtils.ts
```diff
@@ -249,3 +249,51 @@ export function lookupNFTTransfer(transaction: Transaction, tokenId: string): Nf
...
+        const inputResult = ethers.AbiCoder.defaultAbiCoder().decode(functionFragment.inputs, inputArgs)
+        encodedFunction4BytesSignature = inputResult[1].slice(0, 10)
+    } catch (failure) {
+        const f = failure as ethers.EthersError
+        if (f.code === "BUFFER_OVERRUN" || f.code === "INVALID_ARGUMENT") {
+            const inputResult = decodeRedirectForTokenInput(inputArgs)
+            encodedFunction4BytesSignature = inputResult[1].slice(0, 10)
+        }
+    }
+
```
ericleponner, line 272: Why should we attempt calling `AbiCoder.decode()` in this context ?
We know that a special decoding rule is needed ; so let's do it straight.

```

export function resolveFunctionFragmentForHTSProxyContract(functionFragment: ethers.FunctionFragment, inputArgs: string): ethers.FunctionFragment {
    const inputResult = decodeRedirectForTokenInput(inputArgs)
    const encodedFunction4BytesSignature = inputResult[1].slice(0, 10)

    // @notice this is the list of supported method which is collected based on HIP-218 https://hips.hedera.com/hip/hip-218
    const ABI_FOR_SUPPORTED_METHODS = [
...

```

src/schemas/HederaUtils.ts
```diff
@@ -249,3 +249,51 @@ export function lookupNFTTransfer(transaction: Transaction, tokenId: string): Nf
     }
     return result
 }
+
+export function decodeRedirectForTokenInput(inputArgs: string): ethers.Result {
+    const tokenAddress = `0x${inputArgs.slice(2, 42)}`
+    const encodedFunctionSelector = `0x${inputArgs.slice(42)}`
+    return new ethers.Result(tokenAddress, encodedFunctionSelector)
```
ericleponner, line 256: `decodeRedirectForToken() `should check `inputArgs` validity and throw an exception if not.

src/schemas/HederaUtils.ts
ericleponner, line 299: Nice work :)

src/schemas/HederaUtils.ts
quiet-node: I believe inputArgs's validity has already been checked in the wrapper level (https://github.com/hashgraph/hedera-mirror-node-explorer/pull/993/files/1d89d9151b6f07eacd6b777048e5305aa48488cb#diff-968baa6ba16da48193fd76bbf3a04b15eaf0ed55fa5ba73be1173aea08da5c4cR308)

ericleponner (approved):
Looks good.

svienot (approved):
LGTM

svienot (approved):

svienot:
One e2e test is failing for a reason unrelated to these changes -- the search using KNS currently fails.
Thus integrating this PR.

PR #1002, opened 2024-05-01
chore: fixed "The CJS build of Vite's Node API is deprecated" reported by vite

ericleponner:
**Description**:

Since a while, `vite` reports the following deprecation warning:

```
/usr/local/bin/npm run test:unit

> hedera-explorer@0.1.0 test:unit
> vitest run

The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

 RUN  v1.5.3 /Users/eric/Documents/Hedera/WS/hedera-mirror-node-explorer
...
```
Changes below applies [recommended fix](https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated) (`type="module"` in `package.json`).

However this change alone breaks `cypress` execution ;(

```
Your configFile is invalid: /Users/eric/Documents/Hedera/WS/hedera-mirror-node-explorer/cypress.config.ts

It threw an error when required, check the stack trace below:

ReferenceError: exports is not defined in ES module scope
    at file:///Users/eric/Documents/Hedera/WS/hedera-mirror-node-explorer/cypress.config.ts:2:23
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
...
```

So changes also rename `cypress.config.ts` as `cypress.config.js` as recommended [here](https://github.com/cypress-io/cypress/issues/23552#issuecomment-1606338593).


svienot (approved):
LG

PR #1041, opened 2024-05-21
fix: duplication of 'limit' query parameter in axios calls located inside a loop.

svienot:
**Description**:

Almost everything is in the title. We also bump the limit to 100 in these calls to reduce the number of calls to mirror-node since the goal is to get an exhaustive list of entities.

**Notes for reviewer**:

Fix can be observed on mainnet contract: 0.0.5060536


ericleponner (approved):
Looks good.

PR #1122, opened 2024-06-14
fix: `Call Trace` entries contain the same parameters for each call

### LINKED ISSUE #1117: Call stack entries contain the same parameters for each call
### Description

We noticed something in the mirror node for a transaction that seems a bit off.

https://hashscan.io/mainnet/transaction/1718213391.883698003

For each call under call trace, the input parameters are the same.  Its confusing because each call may take different parameters.  The actions endpoint seems to return different input data for each call:

https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x8a89ee664b973de1f3225b8c2118d883a4f64219c202af0a4604629ff2441ddb/actions?limit=100

Each input is a tuple parameter, but only the initial call takes the tuple parameter, not the subsequent calls.

image

### Steps to reproduce

Observe the mirror node call trace for the link above and the actions endpoint to see the difference

### Additional context

_No response_

### Hedera network

mainnet

### Version

whatever version is live

### Operating system

None

ericleponner:
**Description**:

This is a regression introduced in #1000.
Change below restore (wrongly removed) `ContractActionAnalyzer` class and add unit tests to avoid regression.

**Related issue(s)**:

Fixes #1117 

**Notes for reviewer**:

Use [1718213391.883698003](https://hashscan.io/mainnet/transaction/1718213391.883698003) to see fix in action.
See #1117 for additional details.

Fix is deployed on staging server.

**Checklist**

### DIFF
--- a/src/components/contract/ContractActionDetails.vue
+++ b/src/components/contract/ContractActionDetails.vue
@@ -123,9 +123,9 @@ import PlainAmount from "@/components/values/PlainAmount.vue";
 import EVMAddress from "@/components/values/EVMAddress.vue";
 import FunctionInput from "@/components/values/FunctionInput.vue";
 import FunctionResult from "@/components/values/FunctionResult.vue";
-import {ContractResultAnalyzer} from "@/utils/analyzer/ContractResultAnalyzer";
 import FunctionError from "@/components/values/FunctionError.vue";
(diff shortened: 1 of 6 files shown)


svienot (approved):
Looks Good !

PR #1136, opened 2024-06-18
feat: `Search Bar` now uses `api/v1/topics` new REST call when `previewnet` is selected

ericleponner:
**Description**:

Changes below alter `SearchRequest` class so that it uses `api/v1/topics/{topicId}` when `previewnet`.
For `mainnet` and `testnet`, behavior is unchanged (use of `api/v1/topics/{topicId}/messages`).
Changes also include unit tests adaptation to this new logic.

**Related issue(s)**:

Fixes #1109 

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #1137, opened 2024-06-19
chore: explorer now uses `api/v1/topics/{topicId}` for all networks

ericleponner:
**Description**:

Changes below are a follow-up of #1087 and #1136. They
1) remove compatibility code from `TopicDetails` vue and `SearchRequest` class
2) re-activate "should flag deleted topic" test in `TopicDetails` unit tests
3) add missing `TopicIdCache.instance.clear()` in `CacheUtils.clearAll()`
4) remove some unneeded call from `TopicIdCache.load()`

**Checklist**


svienot (approved):
Looks good to me.

PR #1013, opened 2024-05-16
feat: display metadata content and image when available for NFT

svienot:
**Description**:

Modify `NftDetails` view to:
- display image associated to token when available
- make certain metadata attributes prominent when available (name, description, creator)
- display the details of the metadata content when available

**Related issue(s)**:

Fixes #1004

**Notes for reviewer**:

Screenshot 2024-05-16 at 16 46 16

(also on staging server)

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #1149, opened 2024-06-25
chore: follow up to PR #1013

svienot:
**Description**:

One-line change as a follow-up to PR #1013 to make media place-holder appear in NFT tables.


ericleponner (approved):
Looks good.

PR #1208, opened 2024-07-22
feat: add block search capability

### LINKED ISSUE #938: Add Block Number Search Capability
### Problem

The explorer currently lacks the functionality to search for specific block numbers directly from the search bar. Users can search for accounts, transactions, and other entities but cannot query the system using block numbers.

### Solution

To address this gap and improve the utility and user experience of the HashScan Explorer, I propose adding a feature that enables users to search for specific block numbers directly through the search bar.

### Alternatives

_No response_

ericleponner:
**Description**:

Changes below enable to search blocks using their `number` (they could be searched using their `hash` only).

When user now enters number `nnn` in `Search Bar`, Explorer now searches for:
- account, contract, token and topic with id `0.0.nnn`
- block with number `nnn`

In many cases, Explorer finds two matches: an account/contract/token/topic plus a block.
So `Search Bar` has been substantially reworked:
- results matching the searched text are displayed in a dropdown panel
- each result is a navigation link and enable user to make her choice

image

Changes below also include unit tests for new code.

**Related issue(s)**:

Fixes #938 

**Notes for reviewer**:

`Block` search and new `Search Bar` are available on staging server.

**Checklist**


svienot (approved):
Looks good to me!

PR #1239, opened 2024-08-01
chore: pass MouseEvent instead of boolean to RouteManager.routeToXXX() methods.

svienot:
**Description**:

Title says it all.


ericleponner (approved):
Looks good.

PR #1262, opened 2024-08-14
fix: current stake appears in node details when stake for consensus is 0

### LINKED ISSUE #1231: "Current Stake" appears in Node details page in case "Stake for Consensus" is 0
### Description

Exploring the Node details pages I noticed that when the "Stake for Consensus" is 0, a new value "Current Stake" appears in the page.

Also, I cannot find that value looking at the normal Mirror Node queries for [general stake](https://mainnet-public.mirrornode.hedera.com/api/v1/network/stake) and [node details](https://mainnet-public.mirrornode.hedera.com/api/v1/network/nodes), so I'm not sure if it's correct, and if it's correct to show it only when the stake for consensus is 0.

### Steps to reproduce

1. Check https://hashscan.io/mainnet/node/2 and notice "Stake for Consensus" field, and after that the "Min Stake" field.
2. Check https://hashscan.io/mainnet/node/3 and notice "Stake for Consensus" field is 0, and a new "Current Stake" field with value 389,313,830 HBAR appeared.
3. Check any other non-0 staking node (i.e. https://hashscan.io/mainnet/node/4) and the "Current Stake" field disappear again.

### Additional context

01-normal
02-strange

### Hedera network

mainnet

### Version

v24.4.0-172-g40d35e7

### Operating system

None

svienot:
**Description**:

No longer display Current Stake value when Stake for Consensus is 0. Instead show an info tooltip explaining why the value is 0.

**Related issue(s)**:

Fixes #1231 

**Notes for reviewer**:

See screenshot in issue #1231

### DIFF
--- a/src/components/node/NetworkDashboardItem.vue
+++ b/src/components/node/NetworkDashboardItem.vue
@@ -29,8 +29,12 @@
              multiline
              class="h-tooltip">
     <div class="is-flex is-flex-direction-column is-align-items-flex-start">
-      <p v-if="isMediumScreen" class="h-is-property-text mb-1">{{ title }}</p>
-      <p v-else class="h-is-text-size-3 mb-1">{{ title }}</p>
+      <div class="is-flex mb-1">
+        <p :class="{'h-is-property-text':isMediumScreen,'h-is-text-size-3':!isMediumScreen}">
+          {{ title }}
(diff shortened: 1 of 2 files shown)


ericleponner (approved):
Looks good.

PR #1271, opened 2024-08-19
001268 when call target is a token, it should be decoded using dedicated ABIs

### LINKED ISSUE #1268: When call target is a token, it should be decoded using dedicated ABIs (ERC20, ERC721 …)
### Problem

In the following [transaction](https://hashscan.io/testnet/transaction/0x83d02a9f23e65343c8e14c1b1eff4325be7b72703668c52694aeaff56febeeed), `To` field in `Contract Result` section is a [token](https://hashscan.io/testnet/token/0.0.4684974).
Signature `0xa9059cbb` should be interpreted using ERC20 ABI.

### Solution

When call target is a token, the following ABIs are relevant:
- ERC-20 ABI
- ERC-721 ABI
- HIP-206 ABI
- HIP-218 ABI
- HIP-719 ABI

Explorer should embed those ABIs and use them when call target is a token.
`4byte` directory should not be used in that case.

### Alternatives

_No response_

ericleponner:
**Description**:

Changes below update decoding logic applied in `Contract Result`, `Call Trace` and `Event` sections.
When call target (the `To` field) is a token the following ABIs are used:
- ERC-20 and HIP-719 when token is fungible
- ERC-721 and HIP-719 when token is non fungible

Changes also update and add new test cases to validate this new behavior.

**Related issue(s)**:

Fixes #1268

**Notes for reviewer**:

Using the following [transaction](https://hashscan.io/testnet/transaction/0x83d02a9f23e65343c8e14c1b1eff4325be7b72703668c52694aeaff56febeeed), you can see the fix in action:
- without the fix, `Signature` field in `Contract Result `is decoded as `workMyDirefulOwner()` using `4byte` registry (which is incorrect)
- with the fix, `Signature` field is correctly decoded as `transfer()`

**Checklist**


svienot (approved):
Looks great to me!

PR #1297, opened 2024-08-23
chore: refactor table controllers

svienot:
**Description**:

Fixes inconsistencies in the use of local storage keys (for preferred page size of various tables).
Specify missing params (p, k) for TokenRelationshipsTableController.

**Notes for reviewer**:

Causes an unexplained regression in SearchBar e2e test...


ericleponner (approved):
Looks good.

PR #1346, opened 2024-09-10
chore: may be a fix for window undefined error reported by unit tests in CI env

ericleponner:
**Description**:

Changes below add some missing calls to `BalanceAnalyzer.unmount()` in `InlineBalancesValue.spec.ts`.
May be the root cause of the `window undefined` errors that are reported when unit tests are run in GitHub env…

Changes also remove use of the `--sequence.setupFiles` option in `vitest run` (a tentative fix for the pb above … that does not fix the pb above).


svienot (approved):
LG

PR #1432, opened 2024-10-17
fix: missing fungible token balances

### LINKED ISSUE #1413: Missing fungible token balances
### Description

image

### Steps to reproduce

see above

### Additional context

_No response_

### Hedera network

_No response_

### Version

na

### Operating system

None

timi-codes:
**Description**:
This PR fixes missing fungible token balances for accounts with large amount of tokens. It replaces the endpoint `api/v1/balances?account.id={id}` (the call returns a max of 50 token balances per account) with  `/api/v1/accounts/{id}/tokens` to retrieve the token balances of an account. 

**Related issue(s)**:

Fixes #1413 

**Notes for reviewer**:
image

**Checklist**

### DIFF
--- a/src/components/token/TokenCell.vue
+++ b/src/components/token/TokenCell.vue
@@ -48,8 +48,8 @@ import BlobValue from "@/components/values/BlobValue.vue";
 import {TokenInfoCache} from "@/utils/cache/TokenInfoCache";
 import {makeTokenName, makeTokenSymbol} from "@/schemas/HederaUtils";
 import TokenAmount from "@/components/values/TokenAmount.vue";
-import {BalanceCache} from "@/utils/cache/BalanceCache";
 import {TokenType} from "@/schemas/HederaSchemas";
+import { TokenRelationshipCache } from "@/utils/cache/TokenRelationshipCache";
 
 export enum TokenCellItem {
   tokenName = "tokenName",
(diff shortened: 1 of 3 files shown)


svienot:
We can take this fix as a short term relief but let's keep in mind this table will require a deeper refactoring. Using TokenRelationshipCache is not fully adequate as this cache won't retrieve more than 1000 (10*100) tokens, and the number of tokens associated to an account is now unbounded (used to be limited to 1000). So in presence of an account associated with more than 1000 tokens, the issue may arise again.

Would you mind fixing the unit test which fails. I assume it fails because a different axios mock is needed, since we are using a different cache (hence a different underlying REST endpoint).

timi-codes:
Hi @svienot 
I agree that this quick fix works for now. Initially, I considered using the api/v1/accounts/0.0.768084/tokens?token.id=0.0.127877 endpoint to fetch the balance for each token. However, this might not be efficient—especially for accounts with large token holdings, as it would require making a request for each token. For example, an account with 1,000 tokens would need 1,000 API calls to the backend.

That’s why I opted for this temporary solution. One alternative we could consider is modifying the endpoint to return the total number of records. This way, we’d know how many times to call the next endpoint, improving efficiency.

I’m happy to explore any permanent solutions you have in mind.

svienot:
This fix is safe and will already bring a relief. Please go ahead fixing the unit test and we'll integrate the PR. We can revisit this area in a further stage.

timi-codes:
> This fix is safe and will already bring a relief. Please go ahead fixing the unit test and we'll integrate the PR. We can revisit this area in a further stage.

@svienot Thank for the feedback. The unit test has been fixed

svienot (dismissed):
Looks good.

svienot:
@timi-codes - I reviewed and was about to integrate your PR but then was blocked because your commit is not adequately signed. You have a bit of set-up to do, which is [explained here](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).
Basically, you need to generate a GPG Key (if you do not already have one), let GitHub know about it, and configure your IDE for this repo such that your commits are always signed with this key. Then any further commits will show up on GitHub as "Verified" and you will be all set.

timi-codes:
Thank you @svienot, I have integrated a commit signing to my workflow. This and other commits should now be automatically verified.

svienot (approved):
Looks good.

svienot:
@timi-codes - your PR is integrated. Thank you for your contribution!

Issue #1185, opened 2024-07-16
Download transactions button legibility


led-boo:
### Problem

1. Next to the Recent Operations header, the **Download transaction** button is unintuitive due its small size, lack of labelling and no tooltip. Hovering over the icon with a cursor provides no clues and the button must first be noticed and then pressed to identify its function.

<img width="775" alt="Screenshot 2024-07-16 at 11 50 51 am" src="https://github.com/user-attachments/assets/9e667d59-2fc2-46ef-b29f-aa4126087de0">

2. Once pressed, the Download transaction from account window doesn't provide more information as to what is being downloaded or it's format.

<img width="749" alt="Screenshot 2024-07-16 at 12 25 51 pm" src="https://github.com/user-attachments/assets/50aa3577-9860-40e3-afe4-4d51584ade24">

### Solution

1. Increase the size of the Download transaction button with descriptive text so it is more easily identifiable and user friendly. For example; **Download Transaction History**

2. Provide a description as to what is being downloaded i.e format and file type (.CSV) on the Download transactions from account window. For example; _Download the transaction history of account exported as a CSV file. This is useful for record keeping and tax reporting._

mariyaraj:
Hey, can I work on this issue?

hendrikebbers:
@SimiHunjan can you assign @mariyaraj pls? She will work on the issue

Phransis:
Hello, I'm interested in this issue

SimiHunjan:
Hi I assigned @mariyaraj as requested first. If @mariyaraj no longer wants to work on this issue I can assign @Phransis.

hendrikebbers:
She has already created a PR: https://github.com/hashgraph/hedera-mirror-node-explorer/pull/1441

mariyaraj:
Here also the screenshots
![ButtonDownloadTransactionHistory](https://github.com/user-attachments/assets/bee9d601-d9f1-4e54-870e-9ef2c0d7e00f)
![UntertitleAdded](https://github.com/user-attachments/assets/468fd07d-96ea-4485-bbf5-d69173559cb7)

PR #1441, opened 2024-10-17
fix: download Button name and tooltip added

mariyaraj:
Fix for https://github.com/hashgraph/hedera-mirror-node-explorer/issues/1185

### DIFF
--- a/src/dialogs/download/TransactionDownloadDialog.vue
+++ b/src/dialogs/download/TransactionDownloadDialog.vue
@@ -29,6 +29,10 @@
 
     <template #downloadDialogInput>
 
+      <div class="download-hint">
+        This will download the transaction history of the account as a CSV file.
+      </div>
+
       <ContentCell>
         <template #cellTitle>Select Scope:</template>
         <template #cellContent>
@@ -143,5 +147,11 @@ const tokenIdRequired = controller.tokenIdRequired
 <!-- --------------------------------------------------------------------------------------------------------------- -->
 
 <style scoped>
+
+div.download-hint {
+  color: var(--text-secondary);
+  font-size: 12px;
+  font-weight: 400;
+}
 
 </style>
--- a/src/pages/AccountDetails.vue
+++ b/src/pages/AccountDetails.vue
@@ -270,7 +270,7 @@
             <option value="LATEST">LATEST</option>
             <option value="JUMP">JUMP TO DATE</option>
           </SelectView>
-          <DownloadButton @click="transactionDownloadDialogVisible = true"/>
+          <Download :size="24" @click="transactionDownloadDialogVisible = true"/>
           <TransactionFilterSelect v-model:selected-filter="transactionType"/>
         </template>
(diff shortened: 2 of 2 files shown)


ericleponner:
Please join a snapshot to your PR.

mariyaraj:
> Please join a snapshot to your PR.

Could you clarify what kind of snapshot you need? I am new here and I am not sure, what should I do?

ericleponner:
> Could you clarify what kind of snapshot you need? I am new here and I am not sure, what should I do?
#1185 includes two screenshots ; could you include the same screenshots with the visual changes you are proposing.

ericleponner (changes requested):
#1185 includes two screenshots ; could you include the same screenshots with the visual changes you are proposing.

mariyaraj:
Here are screenshots after my changes: 
ButtonDownloadTransactionHistory
UntertitleAdded

src/components/DownloadButton.vue
ericleponner, line 1: - `Download Transaction History` is long : we should replace it by `Download`.
- `Download Transaction History` is fine for `title` attribute but cannot be hard-coded because `DownloadButton` is re-used in other views and different contexts
   - so title / tooltip must be parameterized by adding a new prop to `DownloadButton` (eg `buttonTitle`)
   - some tooltip text must be defined for other `DownloadButton` usage ; your proposals are welcome
- when window width is small, we should hide `Download`and keep the icon only (as it is today)
   - `Footer` view (line 40) shows how to do that

src/components/download/DownloadDialog.vue
ericleponner, line 1: `This is useful for record keeping and tax reporting.` is a bit too much specific : remove it.

src/components/download/TransactionDownloadDialog.vue
ericleponner, line 1: - dialog titles are left-aligned everywhere so keep it left-aligned 
- inline help text should be left-aligned too (we have other dialogs like this)
- remove `This is useful …`
- when possible we try to avoid specific CSS definitions and prefer to use Bulma style classes
    - here you can probably achieve the same visual with `has-text-grey` and `mt-x` classes
    - and remove  `dialog-title-container` and `dialog-subtitle`

ericleponner (changes requested):

mariyaraj:
@ericleponner Changes committed. Here the screenshots:
ButtonDownloadTransactionHistory
DownloadButton
DownloadDialogScreen

svienot:
Rebased on new UI design and adapted to it:
 - the download button does not have any label (per the spec)
 - the download dialog gains a small hint sentence explaining what the action does, so the user can CANCEL if this is not what was expected
 - there is no need for a tooltip on the button, given the above (we avoid profusion of tooltips)
 - replaced the download image by the appropriate Lucide icon

svienot:
Screenshot 2025-02-27 at 18 15 23

ericleponner (approved):
Looks good

ericleponner (approved):

PR #1454, opened 2024-10-23
fix: CLAIM action

svienot:
**Description**:

Fixes 2 issues with CLAIM action:
- reset selected airdrops after 'CLAIM ALL' and 'CANCEL'
- do not display CLAIM button when current airdrop table is empty


ericleponner (approved):
Looks good

ericleponner (approved):

PR #1479, opened 2024-11-05
fix: transaction type filter is broken in Recent Operations section

ericleponner:
Changes below fix a regression in `Recent Operations` table of `Account Details` page.
Transaction type filter became inoperative when enabling `<script setup>` in `AccountDetails.vue`.

**Checklist**


svienot (approved):
Looks good.

PR #1485, opened 2024-11-11
chore: replacing VITE_APP_XXX build time flags by runtime configuration from core-config.json

ericleponner:
**Description**:

Changes below replace `VITE_APP_XXX` build time flags by `core-config.json` configuration file.
Additionally they reorganize `RouteManager` class and replace `NetworkRegistry` by `NetworkConfig`.

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #1496, opened 2024-11-12
chore: update account without staking

svienot:
**Description**:

- follow-up to PR #1485: disable editing properties related to staking when the current network configuration does not have enableStaking = true.
- (unrelated:) fix NodeDetails unit test (abusively copy-pasted from AccountDetails unit test)


ericleponner (approved):
Looks good

PR #1505, opened 2024-11-18
chore: migrate AllowancesSection to <script setup>

svienot:
**Description**:

Changes below migrate `AllowancesSection` component to `<script setup>` syntax.


ericleponner (approved):
Looks good.

Issue #1430, opened 2024-10-16
Rename any usage of “Hedera”, “Swirlds” or “Hashgraph” to “Hiero”


SimiHunjan:
### Problem

This project will be listed under the Linux Foundation Hiero project. This projects needs to rename any rename any usage of “Hedera”, “Swirlds” or “Hashgraph” to “Hiero.” 

### Solution

This projects needs to rename any rename any usage of “Hedera”, “Swirlds” or “Hashgraph” to “Hiero.”

svienot:
Closing this issue since (hopefully) all remaining occurrences of (_Hedera_, _Hashgraph_, _Swirlds_) should now have been renamed to _Hiero_.

The following exceptions remain:

- copyright notices in all source files
- paths for files, repositories, etc...

The assumption is the above exceptions can be addressed at a later stage (but while they remain, it is harder to make sure there is no other occurence left).

PR #1541, opened 2024-12-04
chore: add configuration for cookies acceptance dialog

svienot:
**Description**:

The changes below introduce a new `CoreConfig.cookiesDialogContent` configuration parameter to control the content and appearance of the cookies dialog. 

Note that if not set, the value of the `googleTagID` parameter is ignored.

On the other hand, if `cookiesDialogContent` is set, and `googleTagID` is not, the dialog will be show but will have no other effect (for the time being). This will allow us to use cookies for other features if needed later.

**Related issue(s)**:

Related to #1430


ericleponner (approved):
LGTM

PR #1565, opened 2024-12-13
fix: wrong hbar values assigned to Gas Limit, Gas Used, and Gas Consumed

### LINKED ISSUE #1559: Wrong hbar values assigned to Gas Limit, Gas Used, and Gas Consumed
### Description

The explorer considers the max fee per gas as the gas cost and multiplies it by the gas amounts to report the costs in hbars, but this is wrong.

The maximum fee per gas is set by the user and is ignored by Hedera.

The real cost is defined by [the network fees at the time of the transaction](https://testnet.mirrornode.hedera.com/api/v1/network/fees?order=desc&timestamp=eq%3A1733794756.191981000), which in [this example](https://hashscan.io/testnet/transaction/1733794756.191981000) was 31 at the time of the tx.

So, for example, the correct cost for Gas Consumed in the above example should be 102,297 * 31 = 3171207 = 0.03171207ℏ and not 0.03989583ℏ.

Please note that, due to [this MN's bug](https://github.com/hashgraph/hedera-mirror-node/issues/9942), the current fees value returned by the API is wrong, so the value reported in the fees section (0.03068910ℏ) is actually correct.

### Steps to reproduce

1. Check an ethereum transaction on hashscan (i.e., https://hashscan.io/testnet/transaction/1733794756.191981000)
2. Verify the cost in hbar is derived by the multiplication of the gas by the max fee per gas value

### Additional context

Screenshot 2024-12-11 at 19 47 54

### Hedera network

mainnet, testnet, previewnet

### Version

v24.11.0-29-g5950d0b9

### Operating system

None

svienot:
**Description**:

With the changes below:
- we use script setup syntax in ContractResult and do a tiny bit of refactoring
- we always display the Gas Price in the Contract Result and get its value from `/api/v1/network/fees` at the corresponding timestamp. 

[See detailed analysis of the issue in #1559]

**Related issue(s)**:

Fixes #1559 

**Notes for reviewer**:

Before: 
394906549-1d32f539-08eb-4abe-8776-d63e5fa40f1f

After:
Screenshot 2024-12-13 at 18 15 41

### DIFF
--- a/src/components/contract/ContractResult.vue
+++ b/src/components/contract/ContractResult.vue
@@ -79,7 +79,7 @@
           <template v-slot:value>
             <GasAmount
                 :gas="contractResult?.gas_limit"
-                :price="displayedPrice"
+                :price="gasPrice"
             />
           </template>
         </Property>
@@ -88,7 +88,7 @@
           <template v-slot:value>
             <GasAmount
                 :gas="contractResult?.gas_used"
-                :price="displayedPrice"
+                :price="gasPrice"
             />
           </template>
         </Property>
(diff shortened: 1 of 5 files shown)


ericleponner (approved):
Looks good.

PR #1636, opened 2025-02-18
fix!: Add info-tooltips for Gas Used and Gas Consumed in ContractResult.

### LINKED ISSUE #1621: feat: add help hints to gas used/gas consumed fields on contract executions
### Problem

Contract executions now helpfully show both gas used and gas consumed

Image

But it may not be clear to the typical user what these mean!  (I.e., what's the difference between them).  (In fact, I've had to explain the difference twice  just this month, asked by Hashgraph devs!)

### Solution

I suggest a info popup or tooltip or whatever you call it when there's a little i-in-a-circle and you hover over it and get an explanation.  Or something like that.

Maybe even a link to the [doc page](https://docs.hedera.com/hedera/core-concepts/smart-contracts/gas-and-fees#gas-reservation-and-unused-gas-refund) somehow.

### Alternatives

_No response_

svienot:
**Description**:

All is said in the title.

Note this is in no way a breaking change, but marked with `fix!` in order to trigger the switch to version numbers 25.x.y moving forward.

**Related issue(s)**:

Fixes #1621 

**Notes for reviewer**:
Screenshot 2025-02-18 at 15 56 29
Screenshot 2025-02-18 at 15 56 38

### DIFF
--- a/src/components/contract/ContractResult.vue
+++ b/src/components/contract/ContractResult.vue
@@ -83,7 +83,7 @@
             />
           </template>
         </Property>
-        <Property id="gasUsed">
+        <Property id="gasUsed" :tooltip="gasUsedTooltip">
           <template v-slot:name>Gas Used</template>
           <template v-slot:value>
             <GasAmount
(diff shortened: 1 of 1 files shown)


ericleponner (approved):
Looks good.

PR #1654, opened 2025-02-26
chore: removing unused code from MainDashboard vue

ericleponner:
**Description**:

Changes below remove some unused code: `MainDashboard` vue creates a `TPSMetricLoader` instance which is not used but generates some useless calls to `api/v1/blocks`.

**Related issue(s)**:

None

**Notes for reviewer**:

1) Go https://hashscan.io
2) Open `Network` tab in `Developer` panel
    => without the fix, periodic calls to `api/v1/blocks` are executed
    => with the fix, those calls no longer exist

**Checklist**


svienot (approved):
LGTM

PR #1662, opened 2025-02-27
chore: update ERC Registry with latest ERC-20 and ERC-721 tokens on Hedera mainnet

quiet-node:
**Description**:
This PR updates ERC Registry with latest ERC-20 and ERC-721 tokens on Hedera mainnet


svienot (approved):
LGTM

PR #1668, opened 2025-02-27
chore: add initial erc-1155.json registry files for mainnet and testnet.

svienot:
**Description**:

Including initial version of these such that the automated process from smart-contracts repository can then keep them up-to-date.


ericleponner (approved):
Looks good

PR #1670, opened 2025-02-27
fix:  YTD chart range should be rename 1Y 

ericleponner:
**Description**:
Change below renamed YTD chart range as 1Y.

**Checklist**


svienot (approved):
LGTM

PR #1678, opened 2025-03-03
chore:  different icon in InfoTooltip for warning message

svienot:
**Description**:

Bring back the use of a triangle alert icon (vs. info icon) in InfoTooltip in case of warning message (was lost during UI revamp).


ericleponner (approved):
LG

PR #1703, opened 2025-03-06
chore: exclude SearchBar e2e tests which query name services from CI tests.

svienot:
**Description**:

Extract the tests querying KNS and HNS (as `NameServiceSearch`) and move under `local/` directory to exclude from CI testing.


ericleponner (approved):
LGTM

PR #1709, opened 2025-03-07
chore: configure network colors with builtin color map instead of networks-config.json file.

svienot:
**Description**:

The network colors are now automatically attributed to the configured networks. 
There are 3 color maps available for the time being, so if there are more than 3 networks configured, there is a rotation (network4 gets map1, network5 gets map2, etc...).


ericleponner (approved):
LGTM

PR #1715, opened 2025-03-10
chore: improve sourcify setup and remove private for self signed certificate

svienot:
**Description**:

Make set-up of a local sourcify server (to test contract verification) easier and remove the underlying self-signed certificate
and private key from repository. These are generated as needed.


ericleponner (approved):
Nice automation

PR #1723, opened 2025-03-12
chore: TableController.updateRouteQuery() now updates route only when needed.

ericleponner:
**Description**:

Changes below implement an optimization in `TableController` class which avoids a duplicate update of `vue-router `in some cases (which is slow and can cause visual artifacts). 
`TransactionTableControllerXL` class needs an update to implement the same behavior. 
Unit tests also need update because optimization modifies mirror node queries order.

**Checklist**


svienot (approved):
LGTM

PR #1724, opened 2025-03-12
chore: TableController is now in charge of reading/saving page size from/to local storage

ericleponner:
**Description**:

Changes below refactor table management code so that `TableController` is now in charge of reading/saving page size from/to local storage (in place of `TablePageSize`). Doing this avoids `TableController.pageSize` to be first set to an hard-coded value first and then updated by `TablePageSize` later (which causes useless table state updates).

**Checklist**


svienot (approved):
LGTM !

PR #1751, opened 2025-03-19
feat: support non-zero realms and shards

### LINKED ISSUE #1626: Support non zero realms and shards
### Problem

The explorer needs to support non zero realms and shards in entity IDs. Non zero realms is the primary requirement and if non zero shards can be handled at the same time we can do that. 

The user should be able to search for entity IDs that have non zero realms and shards. Ex: 0.1.2, 1.2.3, etc.

### Solution

Verify we can support non zero shard and realm numbers in entity IDs. 

### Alternatives

_No response_

svienot:
**Description**:

Remove assumptions about realm=0 and shard=0 throughout the code and instead rely on new `baseRealm` and `baseShard` properties in network configuration.

**Related issue(s)**:

Fixes #1626 

**Checklist**


ericleponner (approved):
LG

PR #1755, opened 2025-03-21
chore: Network Fees chart now becomes linear chart and uses gradient and colors from UI design spec

ericleponner:
**Description**:

Changes below turn `Network Fees` chart as a linear chart with the gradient and colors specified in UI design.

image

**Checklist**


svienot (approved):
Looks good !

PR #1756, opened 2025-03-21
chore: ChartController now triggers data load when its HTML container intersects viewport

ericleponner:
**Description**:

`ChartController` now triggers data load only when its HTML container element intersects viewport.
This enables to avoid useless queries to Hgraph data server.

**Checklist**


svienot (approved):
Looks good (and smart) !

PR #1761, opened 2025-03-24
fix: explorer gives a blank page when cookie are fully disabled

ericleponner:
**Description**:

When `cookies` are fully disabled, Explorer crashes at load time and displays a blank page.
Root cause is the following `import` line in `WalletConnectAgent` class:
```
import SignClient from "@walletconnect/sign-client";
```
=> `@walletconnect/sign-client` does not work when `cookies` are fully disabled.

Changes below make sure that Explorer starts correctly whether cookies are disabled or not.
Import line above is replaced by a dynamic import (which also reports an error but at connection time only).
In that case, Explorer reports issue with a specific message:

image

Additional change: when there is no wallet at all, Wallet Chooser dialog now displays the following message (in place of an empty list of wallets):
image

**Checklist**

### DIFF
--- a/src/dialogs/wallet_chooser/WalletChooserDialog.vue
+++ b/src/dialogs/wallet_chooser/WalletChooserDialog.vue
@@ -27,13 +27,23 @@
     <template #modalDialogTitle>Connect Wallet</template>
 
     <template #modalDialogContent>
-      <div style="display: flex; align-items: center; justify-content: center;">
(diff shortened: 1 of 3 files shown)


svienot (approved):
LGTM

ericleponner:
I reported issue to WalletConnect team : see https://github.com/WalletConnect/walletconnect-monorepo/issues/5803.

PR #1765, opened 2025-03-25
fix: fixing pagination in Node Table

ericleponner:
**Description**:

Changes below update pagination setup in `Node Table` view.
This enables to fix the truncated list of  network nodes observed recently.
It seems that, in latest release of Oruga, default value `o-table.per-page` is now 20 (and no longer the number of items in data array).  Fix consists to set `per-page` with item count.

**Checklist**


svienot (approved):
LGTM

PR #1780, opened 2025-03-26
chore: remove unused assets

svienot:
**Description**:

House keeping:
- remove assets no longer used
- gather scripts under ./scripts/


ericleponner (approved):
LG

PR #1782, opened 2025-03-26
chore: remove obsolete setup related to Docker env variables.

svienot:
**Description**:

Remove obsolete setup related to Docker env variables once used to add network configuration to Docker. This is done by means of networks-config.json now (and the Docker config was a hack unusable in K8S environment anyway).


ericleponner (approved):
LGTM

PR #1790, opened 2025-03-27
chore: explorer customisation

svienot:
**Description**:

- Remove `install-branding.sh` script and remove related instructions from the npm targets. The customisation of the explorer for deploying hashscan is now applied from the customisation repository.
- Extract some of the CSS definitions (fonts and theme/color variables) from `explorer.css` and move them to the `public/style-config.css`. The content of this file may be customized at deploy time, similar to `core-config.json` and `networks-config.json`.


ericleponner (approved):
LGTM

PR #1802, opened 2025-04-01
fix: mask HTTP 400 errors in ContractDetails

svienot:
**Description**:

Use a private axios instance in `ERCUtils` to make calls to `api/v1/contracts/call`. These contract calls return CONTRACT_REVERT_EXECUTED when the contract does not support the interface, causing MN to return an HTTP 400 error, which is by default highlighted in the page header by `AxiosMonitor`.


ericleponner (approved):
LG

PR #1887, opened 2025-04-30
feat: improve the summary of parent CONTRACTCALL transactions

### LINKED ISSUE #1644: The HBAR transfer (and tx fee) in the first contract call of a transaction doesn't appear on the 'recent operations' page besides saying 'Contract Call'.
### Problem

I commented this on https://github.com/hashgraph/hedera-mirror-node-explorer/issues/1598, however thought it may be best as it's own issue.

The HBAR transfer (and tx fee) in the first contract call of a transaction doesn't appear on the 'recent operations' page besides saying 'Contract Call'. 
To someone who doesn't know HashScan very well this would be extremely confusing, as initially it can look like they're sending assets and receiving nothing back. You you have to either click on that specific contract call, or go on the transaction details page (where it's still only labelled as 'contract call') and click the parent 0 transaction.

e.g. If you're unfamiliar with SaucerSwap swap or smart contract transactions, this from first glance looks like a user sent 1088 SAUCE somewhere and got nothing in return (https://hashscan.io/mainnet/transactionsById/0.0.7676946-1739368065-776000000). Users have created real support tickets about situations like this in the past. 

Image

### Solution

While keeping the 'Contract Call' type and contract ID; add the hbar transfer info to the 'recent operations' overview, and the transaction overview page, similar to how transfers are displayed for Crypto Transfer transactions. e.g. 

Image

### Alternatives

_No response_

svienot:
**Description**:

When a CONTRACT CALL transaction contains an actual transfer of crypto (i.e. not just fees) we display a transfer graph instead of Contract ID: 0.0.12345 in its transaction summary.

**Related issue(s)**:

Fixes #1644

**Notes for reviewer**:

Before fix:

Screenshot 2025-04-30 at 14 41 50

After fix:

Screenshot 2025-04-30 at 14 42 17

**Checklist**

### DIFF
--- a/tests/unit/Mocks.ts
+++ b/tests/unit/Mocks.ts
@@ -2011,10 +2011,6 @@ export const SAMPLE_BATCH_TRANSACTION: TransactionResponse = {
     }
 }
 
-//
-// https://mainnet-public.mirrornode.hedera.com/api/v1/transactions/0.0.48113503-1662470948-432078184
-//
-
 export const SAMPLE_PARENT_CHILD_TRANSACTIONS: TransactionResponse = {
     "transactions":
         [{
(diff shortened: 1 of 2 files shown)


ericleponner (approved):
Looks good

PR #1920, opened 2025-05-07
chore: bump @oruga-ui/oruga-next to 0.10.5 and @oruga-ui/theme-oruga to 0.6.0.

svienot:
**Description**:

Bump Oruga dependencies to their latest versions -- they need to be synchronised so individual PR from dependabot cannot make it.


ericleponner (approved):
Looks good

PR #1930, opened 2025-05-12
chore: refactor Tooltip component

svienot:
**Description**:

Refactor `Tooltip component` and add a `#content` slot so it can be leveraged instead of using directly o-tooltip when the tooltip content is structured. Apply to NodeTable.


ericleponner (approved):
LG

PR #1937, opened 2025-05-13
chore: refactor ERC1155Cache

svienot:
**Description**:

These changes turn ERC1155Cache into a SingletonCache instead of an EntityCache.


ericleponner (approved):
LG

PR #1970, opened 2025-05-22
chore: re-enabling type checking for oruga-ui components

ericleponner:
**Description**:

Changes below update `oruga-ui` setup in `vuejs` to fix warnings related `o-xxx` components:
- `oruga` is no longer registered globally in `main.ts`
- explicit `imports` are specified in each view that uses `oruga` components

=> this fixes warnings reported everywhere `oruga` components where used
=> this re-enables some level of type checking in matching *`.vue` files
=> some additional changes are required to satisfy re-enabled type checks.

**Checklist**

### DIFF
--- a/src/components/InfoTooltip.vue
+++ b/src/components/InfoTooltip.vue
@@ -37,7 +37,7 @@ const props = defineProps({
     default: null
   },
   position: {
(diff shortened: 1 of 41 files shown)


svienot (approved):
LGTM

PR #1971, opened 2025-05-22
chore: display entity and transaction IDs in page headers

svienot:
**Description**:

In Account/Contract/Token/Topic/Transaction details page, display the ID in the page header title instead of the first card header. 
This allow to:
 - make the ID of the instance examined more prominent
 - free space in the the card header for various labels

**Notes for reviewer**:

Deployed on staging server.


ericleponner (approved):
Looks good

PR #1974, opened 2025-05-23
chore: homogenise action buttons

svienot:
**Description**:

Action buttons (in details pages) are no longer using network color (like default button in dialog).
The only remaining exception is on the Staking page.

Unrelated: remove display of checksum in title of MyStaking section (a label may be displayed in place of the ID here). Furthermore the checksum is useful here since this is by definition the account of the user (account from the connected wallet), so there is no ambiguity.


ericleponner (approved):
LG

PR #2076, opened 2025-06-19
chore: WalletConnectAgent class no longer use async promise executor

ericleponner:
**Description**:

Changes below fix a warning reported by `eslint` when `no-async-promise-executor` flag is on.
They re-write `WalletConnectAgent.waitForApprovalOrModalClose()` to avoid passing an `async` function to `Promise` constructor. And they switch  `no-async-promise-executor` flag on.

**Checklist**

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -25,7 +25,6 @@ export default tseslint.config(
         },
         rules: {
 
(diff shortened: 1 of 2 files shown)


svienot (approved):
LGTM

PR #2102, opened 2025-06-27
fix: interpretation of parent/child relationship between transactions

svienot:
**Description**:

When we get a set of transactions with the same ID, we currently assumes that once we find child (i.e. tx with a non-null parent_consensus_timestamp which are not in a batch), we determine that the parent is the tx with a nonce of 0, and from that we assume that all other transactions are children.

This is wrong since there might be unrelated transactions among this set, i.e. tx with a nonce > 0 and parent_consensus_timestamp === null.

An example of that can be obtained with this query to MN:
https://testnet.mirrornode.hedera.com/api/v1/transactions/0.0.4419734-1722372182-141436565
The first tx returned as a nonce of 1 and a null parent_consensus_timestamp. It is hence not a child of the transaction of nonce 0.

**Notes for reviewer**:

**Before the fix:**
Screenshot 2025-06-27 at 19 47 00
Screenshot 2025-06-27 at 19 49 17

**With the fix:**
Screenshot 2025-06-27 at 19 48 33
Screenshot 2025-06-27 at 19 49 08

**Checklist**

### DIFF
--- a/src/components/transaction/TransactionByIdTable.vue
+++ b/src/components/transaction/TransactionByIdTable.vue
@@ -112,10 +112,12 @@ const makeRelationshipLabel = (row: Transaction): string => {
   } else if (row.scheduled) {
     result = "Scheduled"
   } else if (hasChild.value) {
-    if (row.nonce && row.nonce > 0) {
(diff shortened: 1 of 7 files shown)


ericleponner (approved):
Looks good

PR #2108, opened 2025-06-30
fix: don't encode shard and realm in EVM address

### LINKED ISSUE #2004: EVM address must not encode realm and shard values
### Problem

Previously, realm and shard values were encoded in the EVM address. The new direction is to avoid encoding these values directly in the EVM address - only the EVM address itself should be included. If shard and realm values need to be referenced, they should be passed separately using the following format: shard.realm.evmAddress (e.g., 0.0.evmAddress).

### Solution

Verify if the explorer does any additional encoding of the evm address and if so update it.

### Alternatives

_No response_

svienot:
**Description**:

Do not encode shard/realm in EntityID.toAddress(). 
The resulting EVM address will always have the first 12 bytes set to 0.

**Related issue(s)**:

Fixes #2004

### DIFF
--- a/src/utils/EntityID.ts
+++ b/src/utils/EntityID.ts
@@ -78,8 +78,6 @@ export class EntityID {
         const buffer = new Uint8Array(20);
         const view = new DataView(buffer.buffer, 0, 20);
 
-        view.setInt32(0, this.shard);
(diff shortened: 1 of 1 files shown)


ericleponner (approved):
LG

PR #2407, opened 2026-04-16
fix: incorrect display of hiero hooks storage

svienot:
**Description**:

Fix reactivity in HookStorage to make sure the initial hook id value in the pull-down selector is properly initialized once the slots have been read from MN.

Add unit tests for the HooksSection component.

**Related issue(s)**:

Fixes #2389


ericleponner (approved):
LGTM

PR #2416, opened 2026-04-20
chore: add override for protobufjs.

svienot:
**Description**:

Add override to bump `protobufjs` transitive dependency.


ericleponner (approved):
LGTM

PR #2471, opened 2026-05-12
fix: decoding panic error codes

### LINKED ISSUE #2468: Decode Panic error codes
### Problem

In case of specific errors, smart contracts fail with a `Panic(uint256)` error, for example https://hashscan.io/testnet/transaction/1733163830.893742905/result returns a `Panic(0x32)`.

As a numeric value, it can be hard for developers to understand what the code exactly means.

### Solution

Starting from Solidity v0.8.0, [Panics use specific error codes](https://www.soliditylang.org/blog/2020/12/16/solidity-v0.8.0-release-announcement/#revert-on-assertion-failures-and-similar-conditions-instead-of-using-the-invalid-opcode) to distinguish certain situations in which the panic is triggered.

This is the current list of error codes:

1. 0x01: If you call assert with an argument that evaluates to false.
2. 0x11: If an arithmetic operation results in underflow or overflow outside of an unchecked { ... } block.
3. 0x12: If you divide or divide modulo by zero (e.g. 5 / 0 or 23 % 0).
4. 0x21: If you convert a value that is too big or negative into an enum type.
5. 0x22: If you access a storage byte array that is incorrectly encoded.
6. 0x31: If you call .pop() on an empty array.
7. 0x32: If you access an array, bytesN or an array slice at an out-of-bounds or negative index (i.e. x[i] where i >= x.length or i < 0).
8. 0x41: If you allocate too much memory or create an array that is too large.
9. 0x51: If you call a zero-initialized variable of internal function type.

Hashscan should provide a better description for the error message, something like:

`Panic(0x32) - Accessing an array, bytesN or an array slice at an out-of-bounds or negative index (i.e. x[i] where i >= x.length or i < 0)` instead of just `Panic(0x32)`

### Alternatives

_No response_

ericleponner:
**Description**:

Changes below enable Explorer to decode [Solidity 8.0 panic error codes](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/).

Before
image

After
image

Decoded error codes:
```
const panicMessages = new Map<bigint, string>([
    [0x01n, "assert called with an argument that evaluates to false"],
    [0x11n, "Arithmetic operation results in underflow or overflow outside of an unchecked { ... } block"],
    [0x12n, "Divide or divide modulo by zero"],
    [0x21n, "Convert a value that is too big or negative into an enum type"],
    [0x22n, "Access a storage byte array that is incorrectly encoded"],
    [0x31n, "Call .pop() on an empty array"],
    [0x32n, "Access an array, bytesN or an array slice at an out-of-bounds or negative index"],
    [0x41n, "Allocate too much memory or create an array that is too large"],
    [0x51n, "Call a zero-initialized variable of internal function type"],
])
```
**Related issue(s)**:

Fixes #2468 

**Notes for reviewer**:

Sample: https://hashscan.io/testnet/transaction/1733163830.893742905/result

```
M   _frontend/src/schemas/MirrorNodeUtils.ts
M   _frontend/tests/unit/schemas/MirrorNodeUtils.spec.ts
    # Added fetchSolidityPanicMessage() and matching unit tests

M   _frontend/src/components/values/FunctionError.vue
    # View now calls fetchSolidityPanicMessage() to improve it display

M   _frontend/src/components/contract/ContractResultTable.vue
    # Removed unneeded ?? operator
```

**Checklist**

### DIFF
--- a/_frontend/src/components/contract/ContractResultTable.vue
+++ b/_frontend/src/components/contract/ContractResultTable.vue
@@ -96,7 +96,7 @@ const handleClick = (result: ContractResult, c: unknown, i: number, ci: number, 
 }
 
 const makeErrorMessage = (result: ContractResult) => {
-  return decodeSolidityErrorMessage(result.error_message ?? null)
+  return decodeSolidityErrorMessage(result.error_message)
 }
 
 const results = props.controller.rows
--- a/_frontend/src/components/values/FunctionError.vue
+++ b/_frontend/src/components/values/FunctionError.vue
@@ -37,7 +37,13 @@
     <Property :custom-nb-col-class="customNbColClass" id="errorMessage">
       <template #name>Error Message</template>
       <template #value>
-        <StringValue v-if="decodedError" :string-value="decodedError"/>
+        <span v-if="decodedError">
+          <StringValue :string-value="decodedError"/>
+          <span v-if="decodedPanicMessage !== null">
+            <br/>
+            <span class="h-is-low-contrast h-should-wrap"> {{ decodedPanicMessage }} </span>
+          </span>
+        </span>
         <template v-else>
           <HexaDumpValue :byte-string="error" :show-none="true"/>
           <div v-if="errorDecodingStatus" class="h-is-extra-text">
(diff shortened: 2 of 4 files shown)


svienot (approved):
LGTM

PR #2473, opened 2026-05-12
fix: NFT Images with Special Characters in IPFS URLs

### LINKED ISSUE #2401: Bug: NFT Images with Special Characters in IPFS URLs
### Description

# HashScan Bug: NFT Images with Special Characters in IPFS URLs

## Issue

NFT images fail to load on HashScan when the IPFS URL contains special characters (e.g., `#`). The URL is being passed directly without encoding, so the IPFS gateway can't resolve the file path.

## Example

This IPFS URI:

```
ipfs://bafybeieggepfi5t5lslp6u6etraqamgqavpmmxfumg3rysx5b4fzelpoom/Wookz#525.png
```

Fails because the `#` character is interpreted as a URL fragment instead of being part of the filename. The IPFS gateway returns:

```
Failed to resolve /ipfs/bafybei.../Wookz: no link named "Wookz"
- The root CID exists but the child file does not. - ERR_ID:00020
```

The affected NFT collection is **Wookz (Wookz)**, but this would impact any NFT with special characters in filenames.

## Fix

The file path portion of the IPFS URI needs to be encoded using `encodeURIComponent`, while the CID itself stays as-is:

```javascript
const cid = "bafybeieggepfi5t5lslp6u6etraqamgqavpmmxfumg3rysx5b4fzelpoom/Wookz#525.png";

const encodedCid = cid.split('/').map((part, index) => {
  return index === 0 ? encodeURI(part) : encodeURIComponent(part);
}).join('/');
```

This encodes the filename (`Wookz%23525.png`) so the gateway resolves it correctly, while leaving the CID untouched.

## Source

Reported by a community collaborator who received complaints from NFT holders.

Image

### Steps to reproduce

See details above.

### Additional context

_No response_

### Hedera network

_No response_

### Version

latest

### Operating system

None

svienot:
**Description**:

Refactor and fix blob2URL() utility to ensure that the resulting URL is properly encoded in case it includes special characters.

**Related issue(s)**:

Fixes #2401

**Notes for reviewer**:

Deployed sur stage.
Check with following NFT: {explorer_URL}/mainnet/token/0.0.10033743/3


ericleponner (approved):
LGTM

PR #2478, opened 2026-05-13
chore: override protobufjs version to 8.2.0

svienot:
**Description**:

Set override to 8.2.0 (was 8.0.1 which is vulnerable).


ericleponner (approved):
LGTM

PR #2486, opened 2026-05-13
chore: override systeminformation version to 5.31.6

svienot:
**Description**:

Update override from 5.31.1 (vulnerable) to 5.31.6


svienot:
Dependabot took care of it :-)

Issue #2345, opened 2026-03-19
Add Block Node (HIP-1137) Support to Mirror Node Explorer


Reccetech:
### Objective  
Add support for **HIP-1137 registered service nodes** in the Hiero Mirror Node Explorer, including:
- Block Nodes  
- Mirror Nodes  
- JSON-RPC Relays  

### Scope  
Implement HashScan-style UX:
- Top-level tables:
  - Consensus Nodes (rename existing “Node Table”)
  - Block Nodes
  - Mirror Nodes
  - JSON-RPC Relays
- Detail pages per service node:
  - `/block-node/{registeredNodeId}`
  - `/mirror-node/{registeredNodeId}`
  - `/json-rpc-relay/{registeredNodeId}`

### Table Columns (All Service Types)
- Registered Node ID  
- Description  
- Service Endpoints (first + “+N more”)

### Detail Page Sections
- Overview (ID, Description, Admin Key, Service Type)  
- Service Endpoints (endpoint, port, TLS required)  
- Associated Consensus Nodes  

### Consensus Node Page Updates
- Rename navigation to **Consensus Nodes**

PR #2519, opened 2026-05-22
fix: registered node transaction labels and selector options

svienot:
**Description**:

Follow up to the implementation of issue #2345 (HIP-1137 support). 
Use proper type label and selector options for REGISTERED NODE CREATE/DELETE/UPDATE

Unrelated:
 - change default config of arweave server URL to use arweave.ar.io instead of arweave.net
 - fix interpretation of entity_id in the case of NODE CREATE (account ID of the created node). e.g. below:

Screenshot 2026-05-22 at 13 26 54

**Notes for reviewer**:

Deployed on staging.


ericleponner (approved):
LGTM

PR #2520, opened 2026-05-22
fix: layout of compact TransferGraphSection with mixed types of transfers

### LINKED ISSUE #2239: Transaction content merges together when sending hbar & HTS token together
### Description

Image

example tx id https://hashscan.io/mainnet/transaction/1737729056.539430000

On an accounts operation's page, for transactions where hbar and a HTS token were sent in the same transaction, both parts are confusingly on the same line without a space between them as seen above 

### Steps to reproduce

send hbar & hts in the same transaction, look at the accounts transaction history 

### Additional context

_No response_

### Hedera network

_No response_

### Version

v25.1.1+cf98eb24

### Operating system

None

svienot:
**Description**:

Fix the layout of the compact version of TransferGraphSection (used in the CONTENT column of TransactionTable) when it mixes for instance hbar and tokens transfers.
Wrap the inline-grids in a div to avoid their being displayed on the same line when the table is wide enough.

**Related issue(s)**:

Fixes #2239

**Notes for reviewer**:

**Before**:
Screenshot 2026-05-22 at 16 41 28

**After**:
Screenshot 2026-05-22 at 16 43 10

### DIFF
--- a/_frontend/src/components/transfer_graphs/HbarTransferGraphC.vue
+++ b/_frontend/src/components/transfer_graphs/HbarTransferGraphC.vue
@@ -6,52 +6,54 @@
 
 <template>
 
-  <div v-if="hbarTransferLayout.rowCount >= 1" class="graph-container">
+  <div>
+    <div v-if="hbarTransferLayout.rowCount >= 1" class="graph-container">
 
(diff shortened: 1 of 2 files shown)


ericleponner (approved):
LGTM

PR #2566, opened 2026-06-11
fix: block transactions table with more than 100 transactions

### LINKED ISSUE #2511: Blocks / Transactions table does not display pagination controls
### Description

In same cases, Block Transactions table does not display pagination controls.

Image

### Steps to reproduce

1. Goto https://hashscan.io/mainnet/block/95286477
    => `No. Transactions` is 161
2. Select `Transactions` tab
3. Select 100 per page
    => 100 rows are displayed: OK
    => pagination controls (at the bottom right) are not displayed : KO

### Additional context

_No response_

### Hedera network

_No response_

### Version

v26.0.0

### Operating system

None

svienot:
**Description**:

`TransactionsByBlockCache` now uses block timestamp range (`gte:block.timestamp.from` and `lte:block.timestamp.to`) in the initial MN query, and then `drainTransactions()` with a limit set to the block.count in order to obtain all transactions in the block.

Refactor TransactionsByBlockCache to get rid of lint error related to complexity.

Add tests to `TransactionGroupByBlockCache.spec.ts‎`

**Related issue(s)**:

Fixes #2511

Deployed on staging. 
Check with:  `<server-url>/mainnet/block/95286477`

### DIFF
--- a/_frontend/src/schemas/MirrorNodeUtils.ts
+++ b/_frontend/src/schemas/MirrorNodeUtils.ts
@@ -475,12 +475,7 @@ export async function waitForTransactionRefresh(transactionId: string, attemptIn
 
 export async function drainTransactions(r: TransactionResponse, limit: number): Promise<Transaction[]> {
     let result = r.transactions ?? []
-    // let i = 1
     while (r.links?.next && result.length < limit) {
-        // console.log(`drainTransactions (iteration ${i++})`)
-        // console.log(`  - limit:${limit}`)
-        // console.log(`  - result.length:${result.length}`)
-        // console.log(`  - r.links.next:${r.links.next}`)
         const ar = await axios.get<TransactionResponse>(r.links.next)
         if (ar.data.transactions) {
             result = result.concat(ar.data.transactions)
--- a/_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
+++ b/_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
@@ -18,34 +18,41 @@ export class TransactionGroupByBlockCache extends EntityCache<number, Transactio
     //
 
     protected async load(blockNb: number): Promise<Transaction[] | null> {
-        let result: Transaction[] | null
         try {
-            const block = await BlockByNbCache.instance.lookup(blockNb)
-            if (block?.timestamp?.to && block?.count) {
-                const params = {
-                    limit: Math.min(block.count, 100),
-                    timestamp: ["gte:" + block.timestamp.from, "lte:" + block.timestamp.to]
-                }
-                const response = await axios.get<TransactionResponse>("api/v1/transactions", {params: params})
-                result = await drainTransactions(response.data, block.count)
-                if (result.length !== block.count) {
-                    console.warn(`fetchBlockTransactions only retrieved ${result.length} transactions (expected ${block.count}) for block ${blockNb}`)
-                }
-
-                TransactionByHashCache.instance.updateWithTransactions(result)
-                TransactionByTsCache.instance.updateWithTransactions(result)
-            } else {
-                result = null
-            }
+            return await this.fetchBlockTransactions(blockNb)
         } catch (error) {
             if (axios.isAxiosError(error) && error.response?.status == 404) {
-                result = null
-            } else {
-                throw error
+                return null
             }
+            throw error
+        }
+    }
+
+    private async fetchBlockTransactions(blockNb: number): Promise<Transaction[] | null> {
+        const block = await BlockByNbCache.instance.lookup(blockNb)
+        if (!block?.timestamp || !block.count) {
+            return null
+        }
+        const {from, to} = block.timestamp
+        if (!from || !to) {
+            return null
+        }
+        const params = {
+            limit: Math.min(block.count, 100),
+            timestamp: ["gte:" + from, "lte:" + to]
+        }
+        const response = await axios.get<TransactionResponse>(
+            "api/v1/transactions",
+            {params: params}
+        )
+        const result = await drainTransactions(response.data, block.count)
+        if (result.length !== block.count) {
+            console.warn(`fetchBlockTransactions only retrieved ${result.length} transactions (expected ${block.count}) for block ${blockNb}`)
         }
 
-        return Promise.resolve(result)
+        TransactionByHashCache.instance.updateWithTransactions(result)
+        TransactionByTsCache.instance.updateWithTransactions(result)
+        return result
     }
 }
 
--- a/_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
+++ b/_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
@@ -24,10 +24,14 @@ export class TransactionGroupByBlockCache extends EntityCache<number, Transactio
             if (block?.timestamp?.to && block?.count) {
                 const params = {
                     limit: Math.min(block.count, 100),
-                    timestamp: "lte:" + block.timestamp.to
+                    timestamp: ["gte:" + block.timestamp.from, "lte:" + block.timestamp.to]
                 }
                 const response = await axios.get<TransactionResponse>("api/v1/transactions", {params: params})
-                result = await drainTransactions(response.data, params.limit)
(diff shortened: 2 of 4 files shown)


_frontend/src/schemas/MirrorNodeUtils.ts
```diff
@@ -473,14 +473,21 @@ export async function waitForTransactionRefresh(transactionId: string, attemptIn
...
 
-export async function drainTransactions(r: TransactionResponse, limit: number): Promise<Transaction[]> {
+export async function drainTransactions(r: TransactionResponse, limit: number, perRequest=limit): Promise<Transaction[]> {
     let result = r.transactions ?? []
     // let i = 1
     while (r.links?.next && result.length < limit) {
         // console.log(`drainTransactions (iteration ${i++})`)
         // console.log(`  - limit:${limit}`)
         // console.log(`  - result.length:${result.length}`)
         // console.log(`  - r.links.next:${r.links.next}`)
```
steven-sheehy: I don't think we should keep commented out debugs.

_frontend/src/schemas/MirrorNodeUtils.ts
```diff
@@ -473,14 +473,21 @@ export async function waitForTransactionRefresh(transactionId: string, attemptIn
...
+
+        if (perRequest > limit - result.length) {
```
steven-sheehy: I don't think this is the correct approach. Since you already set an upper lte:block.timestamp.to, you should just continue to paginate if the next link is not null.

_frontend/src/schemas/MirrorNodeUtils.ts
```diff
@@ -473,14 +473,21 @@ export async function waitForTransactionRefresh(transactionId: string, attemptIn
...
+            const nextUrl = new URL(r.links.next, axios.defaults.baseURL);
+            nextUrl.searchParams.set("limit", (limit - result.length).toString())
```
steven-sheehy: There's no need to update the limit. Should use the limit set in the response.

_frontend/src/schemas/MirrorNodeUtils.ts
steven-sheehy, line 1: Any test for this?

steven-sheehy (changes requested):

_frontend/src/schemas/MirrorNodeUtils.ts
svienot: If we just specify `lte:block.timestamp.to` , this won't work, since MN returns the same limit in the response that we gave in the request.
In other words, the initial request for the block we are looking at (the one with 161 tx) is:
`/api/v1/transactions?limit=100&timestamp=lte:1779286595.909841434`
In the response, we get:
`"next": "/api/v1/transactions?limit=100&timestamp=lt:1779286595.553372004"`
which is of no use since there are in fact only 61 left to get for this block.

The only way I see to proceed without updating the limit ourselves is to provide in the initial request both `lte:block.timestamp.to` and `gte:block.timestamp.from` -- i.e. provide twice the query parameter `timestamp` in the URL, which I don't think the spec covers.
 Is this officially supported by MN and what you advise to do?

steven-sheehy: Yes, you need to include both gte and lte using the block's from and to. We support multiple timestamp query parameters. This is supported and advised.

_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
```diff
@@ -17,6 +17,7 @@ export class TransactionGroupByBlockCache extends EntityCache<number, Transactio
     // Cache
     //
 
+    // eslint-disable-next-line complexity
```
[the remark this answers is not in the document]
svienot: Refactored load() to reduce its complexity (which indeed was reported as 10 by eslint).

_frontend/src/schemas/MirrorNodeUtils.ts
svienot: Removed those.

_frontend/src/schemas/MirrorNodeUtils.ts
```diff
@@ -473,14 +473,21 @@ export async function waitForTransactionRefresh(transactionId: string, attemptIn
...
+            r.links.next = nextUrl.pathname + nextUrl.search;
+        }
```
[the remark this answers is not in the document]
svienot: The initial response will never have more than `limit` items.

_frontend/src/schemas/MirrorNodeUtils.ts
svienot: `TransactionGroupByBlockCache` is now refactored such that the initial request is made with both `gte:block.timestamp.from` and `lte:block.timestamp.to` parameters, so we can just iterate while the `next` link is not null.

_frontend/src/schemas/MirrorNodeUtils.ts
svienot: TransactionGroupByBlockCache is now refactored such that the initial request is made with both gte:block.timestamp.from and lte:block.timestamp.to parameters, so we can just iterate while the next link is not null.

_frontend/src/schemas/MirrorNodeUtils.ts
svienot: Tests added in TransactionGroupByBlockCache.spec.ts

_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
```diff
@@ -18,30 +18,41 @@ export class TransactionGroupByBlockCache extends EntityCache<number, Transactio
...
+        const block = await BlockByNbCache.instance.lookup(blockNb)
+        if (!block?.timestamp || !block.count) {
+            return null
+        }
+        const {from, to} = block.timestamp
+        if (!from || !to) {
+            return null
+        }
+        const params = {
+            limit: Math.min(block.count, 100),
```
steven-sheehy, line 41: This should also do min with the user requested page size?  Otherwise hashscan will over-request transactions that the user may not ever request or see. In fact, I don't think you need to min it with 100 or block.count and can purely use the user requested amount.

_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
```diff
@@ -18,30 +18,41 @@ export class TransactionGroupByBlockCache extends EntityCache<number, Transactio
...
+            timestamp: ["gte:" + from, "lte:" + to]
+        }
+        const response = await axios.get<TransactionResponse>(
+            "api/v1/transactions",
+            {params: params}
+        )
+        const result = await drainTransactions(response.data, block.count)
```
steven-sheehy, line 48: Does this mean that we are eagerly loading all transactions in the block when the load the block transactions page? So if there's a block with 20,000 transactions because it was doing 10K TPS it would do 200 calls for the MN even though they might only view the first page? This should be done lazily to avoid overwhelming the MN for large blocks.

steven-sheehy (changes requested):

_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
svienot: Yes, this is currently intentional. At the moment, we hardcode the limit to 100 and load all transactions for a block. As far as I know, this is the only table that works this way, and it has been sufficient so far because blocks rarely contain more than 100 transactions.

We are, however, considering a broader review of how pagination is handled across the explorer. Given that, my preference would be to move forward with this fix now to address the immediate issue of the transaction list being capped at 100 entries, and then tackle any longer-term scalability concerns once we have decided whether and how to redesign pagination more generally.

_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
svienot: See comment below.

_frontend/src/utils/cache/TransactionGroupByBlockCache.ts
svienot: See issue #2580

ericleponner (approved):
LGTM

PR #2658, opened 2026-07-24
feat: support new transaction types

### LINKED ISSUE #2651: Support recently introduced transaction types
### Problem

A number of transaction types are not yet known by explorer, which mean they appear with their raw type as returned by MN API and they do not show up as filtering options in the transaction tables pull down filters.

### Solution

Add those to the list of supported transactions.

### Alternatives

_No response_

svienot:
**Description**:

- Support new transaction types: provide printable form and add them to the transaction filter options
- Remove following types from the filter options (never to be seen seen in transaction table):
   - CRYPTOADDLIVEHASH
   - CRYPTODELETELIVEHASH
- Refactor TransactionFilterSelect (use setup syntax and reduce complexity)
- Update unit tests
- Unrelated: update overrides

**Related issue(s)**:

Fixes #2651


ericleponner (approved):
LGTM

PR #2676, opened 2026-08-04
fix: four bytes analyzer

Neurone:
**Description**

This change fixes the 4-byte signature collision handling for function-call decoding. For transfer-style payloads such as the `a9059cbb` selector, the analyzer now correctly resolves to `transfer(address,uint256)` while preserving the original response ordering from the cache layer.

What changed:
- Adjusted the collision-resolution logic in `frontend/src/utils/analyzer/call/FourByteAnalyzer.ts` to sort a copy of the signature records instead of mutating the original array in place.
- Added regression coverage for:
  - the analyzer’s collision-resolution path in `frontend/tests/unit/utils/analyzer/FourByteAnalyzer.spec.ts`
  - the cache response path in `frontend/tests/unit/utils/cache/SignatureCache.spec.ts`

**Related issue(s)**

Some selectors can return multiple candidate signatures from 4byte.directory. The analyzer should try them in ascending ID order and select the first signature that decodes the call payload correctly. The previous implementation used `Array.sort()` directly on the incoming array, which mutated the input order unexpectedly.

**Notes for reviewer**:

See also https://swirldslabs.slack.com/archives/C0361991PGD/p1785847035673079

**Checklist**

### DIFF
--- a/_frontend/src/utils/analyzer/call/FourByteAnalyzer.ts
+++ b/_frontend/src/utils/analyzer/call/FourByteAnalyzer.ts
@@ -69,7 +69,7 @@ export class FourByteAnalyzer {
 
     private static resolveSignatureCollisions(records: SignatureRecord[], callParams: string): SignatureRecord | null {
         //
-        // Some selectors (like 0x70a08231) have multiple signatures registered on 4bytes.directory.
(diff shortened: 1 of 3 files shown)


ericleponner (approved):
LGTM

svienot:
These changes were integrated via PR #2684

PR #2677, opened 2026-08-05
feat: implement #2247

### LINKED ISSUE #2247: feat: stop using cookies
### Problem

When users visit the explorer for the first time and the `cookiesDialogContent` parameter is configured with a text message, a message appears asking users to accept or decline the use of cookies.

That pop-up gives the impression that users are accepting a standard cookie request, but that is not the case.

Cookies only serve to help the backend provide a better frontend for the user by perhaps remembering previous preferences.

In the case of the explorer, however, what is happening is actually the activation or deactivation of telemetry (via Google Analytics if the corresponding `googleTagID` parameter is correctly configured).

Currently, the explorer writes a cookie only to indicate whether or not the users accepted the message; the information is not used by any backend but is used locally.

In practice, we are talking about a local browser configuration that does not require the use of a cookie (and so any related communication with the users about it) but the explorer still technically uses a cookie for it.

That implies that, when the user **reject** the use of cookies, the explorer still create a cookie saving that rejection, and so the browser sends that info to the backend in the following iterations, actually **enabling** the use of cookies regardless of the user's choice.

### Solution

- Do not use cookies; use local browser configurations via browser storage instead.
- Change the title of the popup from "Accept Cookies" to "Accept Telemetry" to clearly state that the user is accepting telemetry.
- Change the parameter `cookiesDialogContent` to be consistent with its feature (i.e., `telemetryDialogContent`)

Neurone:
**Description**:
This PR:
- implements #2247 
- improves security suggesting npm ci instead of npm install
- makes markdown linter happy

**Related issue(s)**:

Fixes #2247 

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/_frontend/ANALYTICS.md
+++ b/_frontend/ANALYTICS.md
@@ -23,7 +23,7 @@ Explorer reports the following [Google Analytics custom events](https://develope
 > **Note 1**: events above are sent when
 >
 > 1) `VITE_APP_GOOGLE_TAG_ID` is setup in `.env` file
-> 2) User has agreed to telemetry
+> 2) User has agreed to cookies and telemetry
 >
 > **Note 2**: events above are **explicitly** sent by Explorer.
 > `gtag()` function may send other events when [Enhanced Measurement Events](https://support.google.com/analytics/answer/9216061) are enabled in GA account.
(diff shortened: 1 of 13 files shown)


svienot (approved):
LGTM

svienot:
Changes integrated by PR #2685

PR #2684, opened 2026-08-17
fix: 4 byte signature collision handling

### LINKED ISSUE #2687: Fix 4byte.directory signature collision handling
### Description

Some selectors can return multiple candidate signatures from 4byte.directory. The analyzer should try them in ascending ID order and select the first signature that decodes the call payload correctly. The previous implementation used Array.sort() directly on the incoming array, which mutated the input order unexpectedly.

### Steps to reproduce

See also https://swirldslabs.slack.com/archives/C0361991PGD/p1785847035673079

### Additional context

_No response_

### Hedera network

_No response_

### Version

v26.2.0

### Operating system

None

svienot:
**Description**:

This is a shadow PR allowing to integrate work from PR #2676, contributed by @Neurone.

Fixes #2687


ericleponner (approved):
LGTM

PR #2707, opened 2026-08-20
chore: update dependabot grouping

svienot:
**Description**:

Add `oruga` group in `dependabot.yml `to make sure `@oruga-ui/oruga-next` and `@oruga-ui/theme-oruga` are updated in sync.


ericleponner (approved):
LGTM

PR #2713, opened 2026-08-24
feat: customize Oruga to use Lucide icons

### LINKED ISSUE #2686: Customize Oruga to use Lucide icons and remove dependency on Font Awesome icons
### Problem

Oruga is currently configured to use the Font Awesome icon pack whereas the Explorer makes use of Lucide icons everywhere else. 
This is only involved in a couple icons used in the table and pagination components, so it does not justify requiring Font Awesome (and paying for it since we are way beyond its free tier limits).
 Furthermore, our use of Font Awesome is problematic because the way we include it is flagged by security tools.

### Solution

We should:
- Customize Oruga to use a custom icon pack based on Lucide icons (since there is no out-of-the-box support for Lucide in Oruga)
- Remove the dependency on Font Awesome

### Alternatives

_No response_

svienot:
**Description**:

Remove the last dependency on FontAwesome (which is via the icon pack used by Oruga):
 - Configure a custom icon pack for Oruga in `main.ts` based on a custom icon component: `OrugaLucideIcon`
 - `OrugaLucideIcon` is a simple dynamic component which maps icon names as requested by Oruga to Lucide icons.
 
**Note**: `OrugaLucideIcon` is not intended as a generic component mapping any icon used by the Oruga component library, it only deals with the icons used by the Table and Pagination components: `chevron-left`,  `chevron-right`, `loading` (the ultimate goal being to fully remove the Oruga dependency).

**Related issue(s)**:

Fixes #2686

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -57,9 +57,6 @@ updates:
       walletconnect:
         patterns:
           - "@walletconnect/*"
-      fortawesome:
-        patterns:
-          - "@fortawesome/*"
       cypress:
         patterns:
           - "cypress"
--- a/_frontend/index.html
+++ b/_frontend/index.html
@@ -7,7 +7,6 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta property="og:locale" content="en_US">
     <meta property="og:type" content="website">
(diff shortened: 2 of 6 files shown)


ericleponner (approved):
LGTM

PR #2723, opened 2026-08-25
chore: replace mikefarah/yq with step-security/mikefarah-yq

svienot:
**Description**:

Replace github action `mikefarah/yq` with its StepSecurity counterpart: `step-security/mikefarah-yq`.


ericleponner (approved):
LGTM

PR #2726, opened 2026-08-27
ci: tighten workflow token permissions, pin nginx image and add property tests

mgarbs:
**Description**:

Tighten the `GITHUB_TOKEN` permissions in every workflow, pin the nginx base image by digest, and add a small property-based test suite so that the repository scores better on the OpenSSF Scorecard checks that can be addressed from a PR (Token-Permissions, Pinned-Dependencies, Fuzzing). No job loses a scope it uses; write scopes only move from the workflow level to the jobs that need them.

**Changes**

*Token-Permissions*
* `flow-pull-request-checks.yaml`: top level is now `contents: read`; `checks: write`, `pull-requests: write`, `issues: read` move to the three caller jobs (`build`, `unit-tests`, `e2e-tests`), so the reusable workflow keeps receiving exactly what it had before.
* `zxc-compile-explorer-code.yaml`: same set moves from the top level to the `compile` job (`pull-requests: write` is what `vitest-coverage-report-action` uses to comment on the PR).
* `flow-pull-request-formatting.yaml`: `statuses: write` moves to the `title-check` job; top level becomes `contents: read`.
* `flow-release-explorer.yaml`: `id-token: write` / `packages: write` move to the `publish-container` and `publish-chart` caller jobs. `prepare-release` and `finalize-release` keep only the top-level `contents: read`; both jobs authenticate with `secrets.GH_ACCESS_TOKEN` for checkout, `gh` and `semantic-release`, so they never use the built-in token for packages or OIDC.
* `zxc-release-explorer-docker-container.yaml` / `zxc-release-explorer-helm-chart.yaml`: `id-token: write` / `packages: write` move to the `container` / `chart` jobs (ghcr login uses `github.token`).

*Pinned-Dependencies*
* `_frontend/Dockerfile`: `ARG NGINX_TAG` now carries the multi-arch index digest of `nginx:1.27.3` (`sha256:bc2f6a7c8ddbccf55bdb19659ce3b0a92ca6559e86d42677a5a02ef6bda2fcef`, verified against Docker Hub with `docker buildx imagetools inspect nginx:1.27.3`). The same tag is still used, and the build-arg override still works.
* `.github/dependabot.yml`: point the `docker` ecosystem at `/_frontend`, where the Dockerfile actually lives, so the digest gets refreshed by Dependabot instead of going stale.

*Fuzzing*
* Add `fast-check` (4.9.0, exact version like the other devDependencies) and two property-based specs that exercise pure round-trip functions with randomised inputs: `EntityID.parse`/`toString`/`parseWithChecksum`/`toAddress`/`fromAddress`, and `B64Utils` hex, base64 and base32 encode/decode. `package-lock.json` only gains the `fast-check` and `pure-rand` entries.

| Check | Before | After (local `scorecard --local`) |
| --- | --- | --- |
| Token-Permissions | 0 | 10 |
| Pinned-Dependencies | 8 | 9 |
| Fuzzing | 0 | 10 |
| Binary-Artifacts | 10 | 10 |
| Dangerous-Workflow | 10 | 10 |

**Intentionally not changed**
* No CodeQL workflow was added: the repository already runs CodeQL through GitHub's default setup (`dynamic/github-code-scanning/codeql`, "Analyze (actions)" and "Analyze (javascript-typescript)" on every push), which is what gives the SAST check its current score. An advanced-setup workflow cannot upload results while default setup is enabled, so it would only fail.
* `npm install -g semantic-release@... ` in `flow-release-explorer.yaml` is still reported as an unpinned npm command. It installs global tooling with exact versions and there is no lockfile for it; switching to `npm ci` is not applicable there.
* The Cypress `install-command: npm install` input and the Google Chrome `.deb` download in `zxc-compile-explorer-code.yaml` are left as they are.

**Related issue(s)**:

N/A

**Notes for reviewer**:
* `actionlint` on all six workflows reports the same findings before and after (self-hosted runner labels and pre-existing shellcheck style notes only).
* `npm ci` succeeds with the updated lockfile; `npm run type-check` and `eslint` are clean for the new specs; `npm run test:unit` passes apart from `TransactionDownloader.spec.ts > no mock`, which also fails on `main` when run outside UTC (date formatting is timezone dependent) and passes with `TZ=UTC`.
* The one behavioural nuance worth a second look is `prepare-release` / `finalize-release` in `flow-release-explorer.yaml` now running with a read-only built-in token, as described above.

**Checklist**


_frontend/tests/unit/utils/B64Utils.property.spec.ts
ericleponner, line 1: Is there is specific reason to mix unit tests of B64Utils and EntityID  with those ci related changes ?
If none, then I suggest to put those new unit tests in a separate PR.

_frontend/package.json
ericleponner: Why do we need an extra component for unit testing ?
All existing tests have been written without fast-check.

_frontend/tests/unit/utils/B64Utils.property.spec.ts
mgarbs: I can do this

svienot (approved):
LGTM

_frontend/tests/unit/utils/B64Utils.property.spec.ts
svienot: This is a set of changes aimed at significantly improving our [OpenSSF (Open Source Security Foundation) score](https://scorecard.dev/viewer/?uri=github.com%2Fhiero-ledger%2Fhiero-mirror-node-explorer). Let's take them as a single PR. Thanks @mgarbs for contributing this.

mgarbs: To add some context on the why: the new property tests are what satisfies Scorecard's Fuzzing check (it looks for fast-check-based tests in the repo), so they are part of the same scorecard improvement rather than an unrelated addition. Thanks @ericleponner and @svienot for the reviews — happy to put tests in a separate PR next time if that is the preference.

